PR #108, opened 2015-01-18
Update Gemfile

dipeshba:
rb-readline dependency missing as needed in lib/history.rb


deivid-rodriguez:
[Builtin readline](http://ruby-doc.org/stdlib-2.2.0/libdoc/readline/rdoc/Readline.html) is used in `lib/history.rb`, there's not need to use `rb-readline` unless we have a reason.

In any case, it should be added to the gemspec if you wanted it to be a runtime dependency.

PR #113, opened 2015-01-31
Add tests for LocalInterface, remote.rb

eric-hu:
Also changes an instance variable check to silence a warning when running the test suite with "ruby -w" (which is what `rake test` uses)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,9 +10,13 @@ Rake::ExtensionTask.new('byebug', spec) do |ext|
   ext.lib_dir = 'lib/byebug'
 end
 
-desc 'Run the test suite'
+desc 'Run the test suite.  Run a single file with TEST=<file>'
 task :test do
-  files = Dir.glob('test/**/*_test.rb').join(' ')
+  if ENV['TEST']
+    files = ENV['TEST']
+  else
+    files = Dir.glob('test/**/*_test.rb').join(' ')
+  end
   system("ruby -w -Ilib test/test_helper.rb #{files}") || exit(false)
 end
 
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Hi @eric-hu, it's great that your contributions are back!!

Have a look at the build, it's failing. Furthermore, don't feel restricted by the current implementation, refactor whatever needs refactoring. For example, cumbersome tests usually mean refactoring is needed. I though of extracting a `Server` class and a `Client` class inside the `Remote` module, each implementing a `start` method. I even started this refactoring so I could provide you with some diffs if you want.

Thanks again, @eric-hu

eric-hu:
Woops, sorry about the failling build.  I'll take a look into this soon.

Thanks for the clarification about refactoring.  I was definitely a bit
reluctant to refactor the existing code before it was tested, while also
finding a little difficult to test.  I wouldn't mind taking a look at some
diffs for inspiration.  That'd be much appreciated.

deivid-rodriguez:
@eric-hu The only requirement here is that you can still debug remote code after the refactoring, but you can even change the API (`Byebug.wait_connection`, `Byebug.start_server`, `Byebug.start_client`) if you need/want to.

This is the WIP I told you about: https://gist.github.com/deivid-rodriguez/94edb9d8934e98a681b9

Thanks!

deivid-rodriguez:
@eric-hu Did you close this intentionally?

eric-hu:
Woops, I just saw this message buried in my mailbox.  Sorry for the silence!

I closed this (and thought I posted a message) so you didn't have a pull request hanging.  I'd still like to get some more tests in around this before trying a refactor.  I've tried a couple of times to simply refactor, but this code makes some calls to methods/variables defined in C, so a refactor has been hard to work on casually.

deivid-rodriguez:
Sounds good! I don't mind the PR opened, but adding more tests before the refactor sounds like a good idea.

PR #124, opened 2015-02-25
Pass file/line to 'break.created' print

astashov:
Necessary for debugger-xml, since it has to pass that information to IDE


deivid-rodriguez:
Thanks @astashov , sorry I broke that.

So have so tried vim support? Is it working?

astashov:
This thing is missing :)

astashov:
Thanks for merging!

deivid-rodriguez:
:+1:

PR #133, opened 2015-04-01
thread list command fixed to work with debugger-xml

os97673:
the debugger-xml+byebug+RubyMine integration is not complete, but this is the first problem I've found and (I hope) fixed.
For some reasons I was unable to run indent command on Mac
IMHO it would be nice to known what exact version of the utility we are supposed to use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ Gemfile.lock
 lib/byebug/byebug.so
 lib/byebug/byebug.bundle
 .ruby-version
+.ruby-gemset
+.idea
--- a/lib/byebug/commands/thread.rb
+++ b/lib/byebug/commands/thread.rb
@@ -114,7 +114,10 @@ module Byebug
         debug_flag: debug_flag,
         id: context.thnum,
         thread: context.thread.inspect,
-        file_line: file_line || ''
+        file_line: file_line || '',
+        pid: Process.pid,
+        status: context.thread.status,
+        current: (context.thread == Thread.current)
       }
     end
 
--- a/lib/byebug/commands/thread.rb
+++ b/lib/byebug/commands/thread.rb
@@ -52,7 +52,7 @@ module Byebug
         thread_arguments(context)
       end
 
-      print(thread_list)
+      puts(thread_list)
     end
 
     def thread_current(thnum)
--- a/lib/byebug/commands/var.rb
+++ b/lib/byebug/commands/var.rb
@@ -18,7 +18,7 @@ module Byebug
         end
         [v, s]
       end
-      puts prv(vars)
+      puts prv(vars, 'instance')
     end
 
     def var_global(_str = nil)
@@ -41,13 +41,13 @@ module Byebug
       return errmsg(pr('variable.errors.not_module', object: str)) unless is_mod
 
       constants = bb_eval("#{str}.constants")
-      puts prv(constants.sort.map { |c| [c, obj.const_get(c)] })
+      puts prv(constants.sort.map { |c| [c, obj.const_get(c)] }, 'constant')
     end
 
     def var_local(_str = nil)
       _self = @state.context.frame_self(@state.frame)
       locals = @state.context.frame_locals
-      puts prv(locals.keys.sort.map { |k| [k, locals[k]] })
+      puts prv(locals.keys.sort.map { |k| [k, locals[k]] }, 'instance')
     end
 
     def var_all(_str = nil)
--- a/lib/byebug/commands/var.rb
+++ b/lib/byebug/commands/var.rb
@@ -45,8 +45,9 @@ module Byebug
     end
 
     def var_local(_str = nil)
-      _self = @state.context.frame_self(@state.frame)
+      cur_self = @state.context.frame_self(@state.frame)
       locals = @state.context.frame_locals
+      locals[:self] = cur_self unless cur_self.to_s == 'main'
       puts prv(locals.keys.sort.map { |k| [k, locals[k]] }, 'instance')
     end
 
--- a/lib/byebug/printers/plain.rb
+++ b/lib/byebug/printers/plain.rb
@@ -24,7 +24,7 @@ module Byebug
         end
       end
 
-      def print_variables(variables)
+      def print_variables(variables, *_)
         print_collection('variable.variable', variables) do |(key, value), _|
           value = value.nil? ? 'nil' : value.to_s
           if "#{key} = #{value}".size > Setting[:width]
--- a/test/commands/stepping_test.rb
+++ b/test/commands/stepping_test.rb
@@ -27,12 +27,18 @@ module Byebug
 
     def test_next_goes_to_the_next_line
       enter 'next'
-      debug_code(program) { assert_equal 16, state.line }
+      debug_code(program) do
+        assert_equal 16, state.line,
+                     "Unexpected position: #{state.file}:#{state.line}"
+      end
     end
 
     def test_n_goes_to_the_next_line
       enter 'n'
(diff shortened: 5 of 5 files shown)


lib/byebug/commands/threads.rb
```diff
@@ -66,6 +58,29 @@ def parse_thread_num_for_cmd(subcmd, arg)
...
+    end
+
+    def file_line(context, should_show_top_frame)
+      return '' unless should_show_top_frame
+      location = thread_location(context)
+      if location
+        "#{location.path}:#{location.lineno}"
+      else
+        ''
+      end
```
deivid-rodriguez: For consistency with the rest of the code and with the first line of this method, can you change this to use a guard clause (`return '' unless location`).

deivid-rodriguez:
@os97673 Thanks a lot for this PR, looks great.

I've added some comments inline and in a specific commit.

Regarding the `indent` utility, I definitely need to make this clearer in CONTRIBUTING.md. However, I have no idea how to run it on a Mac. If you want to investigate it, that's great. Otherwise we can just skip the task that uses it for Mac contributors.

lib/byebug/commands/threads.rb
os97673: done

os97673:
> I don't get this commit, could you elaborate on why it's necessary?

FinishTestCase#test_finish_does_not_stop_in_byebug_internal_frames and BasicSteppingTestCase#test_next_does_not_stop_at_byebug_internal_frames consider byebug/test/support to be internal, while ThreadTestCase#test_thread_list_marks_current_thread_with_a_plus_sign wants to stop at file which is under byebug/test.
Before my changes this was achieved by file/lineno from regular state, but now we need to calculate the file/lineno using Context#ignored_files.  If we will ignore all files then the latter test will report one of minitest's source files, if we will treat byebug/test as regular app.code (which is not ignored) then the former tests will stop at "internal" frames.
Perhaps there is a better way to resolve the problem, but this is what I've found on my own.

os97673:
> I've added some comments inline and in a specific commit.

fixed/answered

> Regarding the indent utility, I definitely need to make this clearer in CONTRIBUTING.md. However, I have no idea how to run it on a Mac. If you want to investigate it, that's great. Otherwise we can just skip the task that uses it for Mac contributors.

Will do this when/if I will need to change native code.

deivid-rodriguez:
I'll review and merge this tomorrow, thanks a lot Oleg!

deivid-rodriguez:
I'm reviewing this and I'm not totally convinced this is the way to go. What if you remove the line `self.allow_in_control = true` from the `thread list` command? Does it solve the problem?

os97673:
> I'm reviewing this and I'm not totally convinced this is the way to go. What if you remove the line self.allow_in_control = true from the thread list command? Does it solve the problem?

Well, it would solve the problem, but the command has allow_in_control in both debugger and ruby-debug-ide(https://github.com/ruby-debug/ruby-debug-ide/blob/master/lib/ruby-debug-ide/commands/threads.rb#L3) and some IDEs (it looks like RubyMine is not in the list, but I could be missing something) may send it w/o stoping at any breakpoint.  Thus I'd prefer to not change this (debugger-xml is supposed to be replacer for ruby-debug-ide and allow all IDEs who use the latter to use it instead)

deivid-rodriguez:
Well, RubyMine is the first IDE we support so if RubyMine is not in the list and it works, I'd prefer to go with this simpler option.

deivid-rodriguez:
I pushed some stuff to master, so you need to rebase the PR. The changes in the `thread list` command should be minimal if we apply my solution so the rebase shouldn't be too painful (sorry about that, in any case).

deivid-rodriguez:
Also my changes require a minor update in `byebug/debugger-xml`: remove the `should_saw_top_frame` parameter. I can PR that once we get to a decision about  `byebug-xml` vs `debugger-xml`.

deivid-rodriguez:
@os97673 I also think it could be a good idea to remove every monkey patching of byebug's commands from `debugger-xml`. For example, in the case of the thread command I can include the thread `pids` and then ignore them in byebug's printer. This way we will more easily prevent future breakages. What do you think?

os97673:
I will rebase the PR, change debugger-xml according your changes (because I will re-test the changes)
I will also try to simplify the changes by removing "allow_in_control" and let you know if this will not work.
As for removing monkey-patching from debugger-xml I will be happy to do this, but let do it in separate PR.

os97673:
Done, I've updated debugger-xml to work with the recent byebug's changes, RubyMine works ok with current version of the "thread list" command.

deivid-rodriguez:
You mean no changes in `thread list` are required? If so, can you remove the changes in the tests?

os97673:
> You mean no changes in thread list are required?

almost none (just replaced "print" with "puts")

> If so, can you remove the changes in the tests?

Done.  Though I wonder what is the purpose of Context#ignored_file? if you stop at breakpoint in ignored file :(

deivid-rodriguez:
Ok, finally let's clean this up and squash it into the following commits:
* New gitignores (c87e14a8 + 4ec7351)
* `var local` should also report self (8d806e5)
* Specify kinds of variables (3d79dff + f0ca2f5)
* `thread list` should support printers (bfa524c)
* New thread arguments (a6396c4 + 5eb9683)

f712984 and 5f1c824 are the inverse of each other so can be both removed.

> Though I wonder what is the purpose of Context#ignored_file? if you stop at breakpoint in ignored file

That change doesn't change any behaviour, it just changes the tests. I don't think the current code is stopping at ignored files, that's the purpose of `Context#ignored_file` (https://github.com/deivid-rodriguez/byebug/blob/master/lib/byebug/context.rb#L85-95). I think what you were running into was a problem with calculating the first non-ignored frame and decide the current frame location without any context information (I've always found that `drop_while`, `take_while` play with the backtrace very prone to errors).

Also, I've implemented a new command & subcommand API that should make things easier. It should help us removing monkey patches from `debugger-xml` and make them unnecessary in the future. I'll be committing it after we merge this in.

os97673:
done

deivid-rodriguez:
Thanks, now we are almost ready to go, but there's something going on with the _appveyor_ build...

Your PR seems to be consistently failing in the same test (which is a test for really basic functionality of the debugger). I've restarted the build for current master and it seems to consistently pass (just tried again and I'm waiting for the result), so I'd like to investigate what's going on... Any ideas? Do you have a windows machine to look into this?

os97673:
Well, it is an interesting question :) Have no ideas :(  Will try to debug the problem on windows machine at work but it will not happen until Tuesday.

deivid-rodriguez:
No problem Oleg! Let's see what you find out on Tuesday.

os97673:
It looks like the box is busy today, thus I will try to do some investigation using appveyor.
I's trying to configure appveyor for my repo, but in a mean time I've added more informative message to the test.

os97673:
@deivid-rodriguez could you please re-start the latest build because it failed due to timeout when it tries to install gems.
Also is there any special hints on how to setup build on appveyor? In my case it complains: "Specify a project or solution file. The directory does not contain a project or solution file."

deivid-rodriguez:
I restarted the build.

I think you have to modify appveyor.yml to build against your branch.

    branches:
      only:
        - master
        - ide-support

deivid-rodriguez:
It passed now, funny. Let me restart it again.

os97673:
> It passed now, funny. Let me restart it again.

it is sad, but I'd say that you have blinking tests :(

os97673:
Great!!! Thanks.

PR #134, opened 2015-04-08
regexp for VarCommand changed to accept "var inspect <expr>" provided by...

os97673:
... debugger-xml

Need to pass nil instead of empty string to var_xxx

debugger-xml provides "var inspect" command which is used by IDE to evaluate any expression, and so 
it could be something like "var inspect 1 + 1"  (not necessary just one word after "var inspect" prefix.
Thus I've changed the regexp and the way we use it.

### DIFF
--- a/lib/byebug/commands/var.rb
+++ b/lib/byebug/commands/var.rb
@@ -75,7 +75,7 @@ module Byebug
     end
 
     def regexp
-      /^\s* v(?:ar)? (?: \s+(\S+) (?:\s(\S+))? )? \s*$/x
+      /^\s* v(?:ar)? (?: \s+(\S+))?/x
     end
 
     def execute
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hi Oleg, I'm adding a Subcommand API for plugins to use. To add a subcommands, you'll need to do reopen the command class and add another command class inside (implementing a `regexp` and `execute` methods)

    module Byebug
      class VarCommand < Command
        class InspectSubcommand < Command
          def regexp
            /^\s* insp(?:ect)? (?:\s+ (.+))? \s*$/x
          end

          def execute
            # Your custom subcomand's code
          end
        end
      end
    end

My patches to support that include exactly this change, so for convenience I won't be merging this.

os97673:
ok, please let me know when the patch will be pushed to the repo.

deivid-rodriguez:
@os97673 I pushed the stuff I talked to you about. Let me know of any issues, questions or whatever.

os97673:
I've changed debugger-xml and restored changes which were lost during the rearchitecture (#139)

PR #135, opened 2015-04-08
No more monkey patching

os97673:
I've added some information debugger-xml needs from byebug to remove monkey-patching from debugger-xml.
I've made a small change in var.rb which will not allow us to remove monkey-patching completely because we do provide additional sub-commands and replace couple methods (see https://github.com/astashov/debugger-xml/blob/master/lib/byebug/commands/variables.rb)  Any ideas how to improve that is welcome.

Also debugger-xml monkey-patches several commands to make them "unsupported" (e.g. https://github.com/astashov/debugger-xml/blob/master/lib/byebug/commands/help.rb)
Again any ideas how to fix this w/o monkey-patching are welcome ;)

### DIFF
--- a/lib/byebug/commands/eval.rb
+++ b/lib/byebug/commands/eval.rb
@@ -54,8 +54,6 @@ module Byebug
   class EvalCommand < Command
     include EvalFunctions
 
-    self.allow_in_control = true
-
     def match(input)
       @input = input
       super
--- a/lib/byebug/commands/frame.rb
+++ b/lib/byebug/commands/frame.rb
@@ -54,12 +54,14 @@ module Byebug
 
     def get_pr_arguments(frame_no)
       file = @state.frame_file(frame_no)
+      full_path = File.expand_path(file)
       line = @state.frame_line(frame_no)
       call = @state.frame_call(frame_no)
       mark = @state.frame_mark(frame_no)
       pos = @state.frame_pos(frame_no)
 
-      { mark: mark, pos: pos, call: call, file: file, line: line }
+      { mark: mark, pos: pos, call: call, file: file,
+        line: line, full_path: full_path }
     end
   end
 


deivid-rodriguez:
Thanks Oleg. I just added one inline comment.

Regarding removing monkeypatches. I've addressed the first issue with subcommands, I'll push the changes soon. I'm still thinking of the best way to support disabling commands. Maybe a whitelist of supported commands in the printer?

os97673:
> I'll push the changes soon. I'm still thinking of the best way to support disabling commands. Maybe a whitelist of supported commands in the printer?

Well, I'd prefer to have some API in Byebug class for this.  E.g. modifiable list of commands or something similar.  Imho, printer is a wrong place for filtering commands.

deivid-rodriguez:
Yep, I agree. But the idea of a whitelist works?

os97673:
> But the idea of a whitelist works?

do you mean if it is ok to be able to whitelist/blacklist some commands?
If so, then yes, I think it is.

deivid-rodriguez:
> do you mean if it is ok to be able to whitelist/blacklist some commands?
If so, then yes, I think it is.

Sorry I didn't answer this. Yes, I meant that plugins should provided a whitelist of the core byebug commands they want enabled.

I don't currently have time to work on this, just wanted to let you know in case you are waiting for me to move this forward.

Regards.

os97673:
> Sorry I didn't answer this. Yes, I meant that plugins should provided a whitelist of the core byebug commands they want enabled.

Current approach debugger-xml uses is more like black-listing some commands most likely because most of commands you provide are useful for vim integration (for IDE we do not need to blacklist anything because user has not access to the debugger's commands ;)
Thus if/when you will have time for this I'd vote for blacklisting.

deivid-rodriguez:
Got it, makes sense.

PR #137, opened 2015-04-12
Add test coverage to script interface test

eric-hu:
Just adding some test coverage to an untested file.

### DIFF
--- a/test/interfaces/script_interface_test.rb
+++ b/test/interfaces/script_interface_test.rb
@@ -1,0 +1,55 @@
+require 'byebug/runner'
+require 'mocha/mini_test'
+
+module Byebug
+  class ScriptInterfaceTest < MiniTest::Test
+    class InitializeTest < ScriptInterfaceTest
+      def setup
+        @file = mock
+        File.expects(:open).with('file').returns(@file)
+      end
+
+      def test_initialize_wires_up_dependencies
+        interface = ScriptInterface.new('file')
+        assert_equal @file, interface.input
+        assert interface.output.instance_of?(StringIO)
+        assert interface.error.instance_of?(StringIO)
+      end
+
+      def test_initialize_verbose_writes_to_stdout_and_stderr
+        interface = ScriptInterface.new('file', true)
+        assert_equal @file, interface.input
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I'd prefer an integration test to get this covered (load byebug some commands in an rc file), but this is helpful too. Thanks!

eric-hu:
I'm starting to see how integration testing is done in some other files.  I was browsing with someone else the CommandProcessorTest today--is that a good example?

I understand the appeal of integration tests, I'd like to maybe separate out unit and integration tests.  My experience with CommandProcessorTest was that it took a bit of close reading and reasoning to understand what's going on, which makes it harder to jump into.  On the other hand, given the way the test is written, it seems very resilient to refactoring, which is definitely a good thing.  How would you feel about lumping the tests into test/unit/ and test/integration/ ?

deivid-rodriguez:
@eric-hu I consider most of the tests in `byebug` as integration tests, because they simulate the behaviour of an end user and make assertions on the output/results. They involve different classes, modules and so on. Whereas I use unit tests all the time, this kind of tests have proven very effective for `byebug` so I'd rather keep it like this. Not opposed to unit tests like this one, but I favor integration tests here.

What did you find difficult to understand about the `CommandProcessorTest`? It requires you to understand how the test suite works but I don't see how it is different from other tests in `test/commands` (apart from testing behavior not specific to any command).

eric-hu:
Sorry for the late response.  I let this get buried in my inbox.  Hopefully this answer might still be of interest to you.

When I looked through `CommandProcessorTest` and `CommandProcessor`, I didn't know which commands entered in by the user mapped to CommandProcessor.  In order to understand why `n;n` entered into one of the tests in `CommandProcessorTest` covers behavior `CommandProcessor`, I have to read additional files that connect `CommandProcessor` to the rest of Byebug.  The answer was in `CommandList`, but I missed that when I was working on this PR.

I suppose I'm used to a testing style that splits testing into unit and integration tests, with more coverage coming from unit tests.  I wrongly assumed that the tests named {filename}Test were unit tests for {filename}.

PR #139, opened 2015-04-12
restoring changes which were lost in the latest rewrite/merge

os97673:
it looks like most of my recent changes were lost during commands' re-architecture :(
Restoring them.
Do you think we should add some tests to void such problem in future, or you do not expect to do this again?

### DIFF
--- a/lib/byebug/commands/var/const.rb
+++ b/lib/byebug/commands/var/const.rb
@@ -19,7 +19,7 @@ module Byebug
         end
 
         constants = bb_eval("#{str_obj}.constants")
-        puts prv(constants.sort.map { |c| [c, obj.const_get(c)] })
+        puts prv(constants.sort.map { |c| [c, obj.const_get(c)] }, 'constant')
       end
 
       def short_description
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
> it looks like most of my recent changes were lost during commands' re-architecture :(

Sorry about that Oleg, my bad. I rebased my re-architecture on top of your changes, and only reviewed the places where I got conflicts. Given that the re-architecture included some file renames and complete rewrites of commands, your changes were automatically "swallowed" by the rebase without any conflicts.

> Do you think we should add some tests to void such problem in future, or you do not expect to do this again?

I have no idea whether I'll perform this kind of big changes in the future, I'll just do what I think it's best in every case. In any case, tests for any changes in the code are always welcome (do you want me to merge this now or do you want to add the tests here?).

The appveyor build is failing again, but it seems like the same problem as in the previous build and I don't think it's related to this changes.

os97673:
> I have no idea whether I'll perform this kind of big changes in the future, I'll just do what I think it's best in every case. In any case, tests for any changes in the code are always welcome (do you want me to merge this now or do you want to add the tests here?).

I think it is ok to merge the changes now (w/o tests).  I will think how to test the changes.

PR #151, opened 2015-06-26
Fix the use of both control and normal connections

rhapsodyv:
It seems just a typo in this commit: https://github.com/deivid-rodriguez/byebug/commit/90cdfcc9cee8ba6362bad2185390cbe4fa85845d

Using the same var in the two places kills one connect when another is opened. Detected the problem while using sublime-ruby-debugger.

I just changed to use a local var (igual was before that commit), and it fixes everething for me.


deivid-rodriguez:
Thanks!

Issue #144, opened 2015-05-18
When use `byebug` and `continue` once, Ruby process get slow.


kntmrkm:
After use `byebug` and `continue`.
Page load time get slow. Webrick's ruby process got slow maybe.

Before use `byebug`
![before](https://cloud.githubusercontent.com/assets/2703486/7674479/d1ff0906-fd63-11e4-892f-f2d13150edaa.jpg)

After use
![after](https://cloud.githubusercontent.com/assets/2703486/7674481/de7e25f4-fd63-11e4-968b-811d2b6c6b0b.jpg)

Gemfile is just 

```
source 'https://rubygems.org'

gem 'pry-rails'

group :development do
  gem 'rack-mini-profiler'
end

group :test do
  gem 'sqlite3'
end

group :development, :test do
  gem 'pry-byebug'
end

# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '4.2.1'
gem 'pg'
gem 'sass-rails', '~> 5.0'
gem 'uglifier', '>= 1.3.0'
gem 'coffee-rails', '~> 4.1.0'
gem 'jquery-rails'
gem 'turbolinks'
gem 'jbuilder', '~> 2.0'
gem 'sdoc', '~> 0.4.0', group: :doc

group :development, :test do
  gem 'byebug'
  gem 'web-console', '~> 2.0'
  #gem 'spring'
end
```

I've open issue at `pry-byebug`, but This relate to `byebug` maybe.
This is linked this issue.

https://github.com/deivid-rodriguez/pry-byebug/issues/61#issuecomment-102660264

deivid-rodriguez:
@kntmrkm Cool, thanks.

I had an idea on how to improve this. I'll disable the debugger when user hits `continue` and there's no enabled breakpoints. I'll let you know when I get to it, or you can give it a try if you want. :smile:

jhirbour:
We're also experiencing this issue. We are using puma, but otherwise basically the same setup and symptom.

choltz-i4cp:
Same problem here. Page load time increase substantially after debugger is hit. If I remove the byebug call from the code and refresh the page, it still runs slow. Only way to resolve is to stop and restart rails.

kirichkov:
I've also been seeing the behavior described by @choltz-i4cp . I've tried byebug 3.1.2, 4.0.5 and even 5.0 but the bug is present in all of these versions.

why-el:
[...]

PR #160, opened 2015-08-08
Disable tracepoints after continue

k0kubun:
Fixes https://github.com/deivid-rodriguez/byebug/issues/144.

I couldn't find out how to avoid SEGV in tests and what a correct fix would be. I'm just sharing my idea.
`Byebug.start` enables tracepoints and they are left enabled after `continue`. If tracepoints are enabled, it gets slow. In fact, this branch does not change things slow after `continue`. I want to disable them while I'm not debugging.

I think we can disable tracepoints after `continue` or ^D and enable them again when `byebug` or `debugger` is called. Is there any reason to keep tracepoints enabled all the time?

### DIFF
--- a/ext/byebug/byebug.c
+++ b/ext/byebug/byebug.c
@@ -13,6 +13,7 @@ static VALUE tracepoints = Qnil;
 static VALUE raised_exception = Qnil;
 
 static ID idPuts;
+static ID idEmpty;
 
 /* Hash table with active threads and their associated contexts */
 VALUE threads = Qnil;
@@ -585,6 +586,36 @@ Stop(VALUE self)
   return Qtrue;
 }
 
+static VALUE
+Stoppable(VALUE self)
+{
+  VALUE context;
+  debug_context_t *dc;
+
+  if (!IS_STARTED)
+    return Qfalse;
+
+  if (breakpoints != Qnil && rb_funcall(breakpoints, idEmpty, 0) == Qfalse)
+    return Qfalse;
+
+  if (catchpoints != Qnil && rb_funcall(catchpoints, idEmpty, 0) == Qfalse)
+    return Qfalse;
+
+  if (post_mortem == Qtrue)
+    return Qfalse;
+
+  context = Current_context(self);
+  if (context != Qnil)
+  {
+    Data_Get_Struct(context, debug_context_t, dc);
+
+    if (dc->steps > 0)
+      return Qfalse;
+  }
+
+  return Qtrue;
+}
+
 /*
  *  call-seq:
  *    Byebug.start -> bool
@@ -748,6 +779,7 @@ Init_byebug()
   rb_define_module_function(mByebug, "start", Start, 0);
   rb_define_module_function(mByebug, "started?", Started, 0);
   rb_define_module_function(mByebug, "stop", Stop, 0);
+  rb_define_module_function(mByebug, "stoppable?", Stoppable, 0);
   rb_define_module_function(mByebug, "thread_context", Thread_context, 1);
   rb_define_module_function(mByebug, "tracing?", Tracing, 0);
   rb_define_module_function(mByebug, "tracing=", Set_tracing, 1);
@@ -763,4 +795,5 @@ Init_byebug()
   rb_global_variable(&threads);
 
   idPuts = rb_intern("puts");
+  idEmpty = rb_intern("empty?");
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 ### Fixed
 * [#177](https://github.com/deivid-rodriguez/byebug/issues/177). Some issues
 with formatting results of evaluations.
+* [#144](https://github.com/deivid-rodriguez/byebug/issues/144). Ruby process
+after using byebug does no longer get slow.
 * [#121](https://github.com/deivid-rodriguez/byebug/issues/121). `byebug`
 commands inside code evaluated from debugger's prompt are now properly working.
 * Another evaluation bug in autocommands.
--- a/lib/byebug/commands/continue.rb
+++ b/lib/byebug/commands/continue.rb
@@ -41,6 +41,8 @@ module Byebug
       end
 
       processor.proceed!
+
+      Byebug.stop if Byebug.stoppable?
     end
   end
 end
--- a/lib/byebug/interfaces/local_interface.rb
+++ b/lib/byebug/interfaces/local_interface.rb
@@ -3,6 +3,8 @@ module Byebug
   # Interface class for standard byebug use.
   #
   class LocalInterface < Interface
+    EOF_ALIAS = 'continue'
+
     def initialize
       super()
       @input = STDIN
@@ -13,12 +15,12 @@ module Byebug
     #
     # Reads a single line of input using Readline. If Ctrl-C is pressed in the
     # middle of input, the line is reset to only the prompt and we ask for input
-    # again.
+    # again. If Ctrl-D is pressed, it returns "continue".
     #
     # @param prompt Prompt to be displayed.
     #
     def readline(prompt)
-      Readline.readline(prompt, false)
+      Readline.readline(prompt, false) || EOF_ALIAS
     rescue Interrupt
       puts('^C')
       retry
--- a/test/commands/continue_test.rb
+++ b/test/commands/continue_test.rb
@@ -32,6 +32,12 @@ module Byebug
       debug_code(program) { assert_program_finished }
     end
 
+    def test_stops_byebug_after_continue
+      enter 'continue'
+
+      debug_code(program) { assert_equal false, Byebug.started? }
+    end
+
     def test_continues_up_to_breakpoint_if_no_line_specified
       enter 'break 14', 'continue'
 
--- a/test/interfaces/local_interface_test.rb
+++ b/test/interfaces/local_interface_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+require 'minitest/mock'
+
+module Byebug
+  #
+  # Tests the local interface
+  #
+  class LocalInterfaceTest < TestCase
+    def test_continues_by_control_d
+      # `Readline.readline` returns nil for Control-D
+      Readline.stub(:readline, nil) do
+        interface = LocalInterface.new
+        assert_equal 'continue', interface.readline('(byebug)')
+      end
+    end
+  end
+end


deivid-rodriguez:
@k0kubun Thanks! This is the same idea I shared in #144. I agree with you, there's no reason to leave them enabled.

There's two problems with the patch:

* The obvious one: it breaks the tests.
* We cannot disable tracepoints if there are any enabled breakpoints or catchpoints, so an extra check is needed somewhere.

Feel free to further research on those issues. Otherwise, I'll pick it up where you left it when I get to this.

Thanks again!

k0kubun:
Thank you for your quick response!

> there's no reason to leave them enabled.
> We cannot disable tracepoints if there are any enabled breakpoints or catchpoints, so an extra check is needed somewhere.

I created this pull request to hear these points. I'll research further. Thanks.

deivid-rodriguez:
:+1: This will be a killer speed improvement if we get it working! :smiley:

deivid-rodriguez:
@k0kubun Did you have the chance to investigate further?

k0kubun:
I strongly want this change but currently I still coudn't make time to work on this. I have will to investigate this further.

deivid-rodriguez:
Thanks! We are in the same situation. :)

k0kubun:
> We cannot disable tracepoints if there are any enabled breakpoints or catchpoints, so an extra check is needed somewhere.

- I understood it and fixed so in fc1eb51ef90b5a6238fe254d323cf34b3ef03b68.

> There's two problems with the patch:
> The obvious one: it breaks the tests.

- I recognized sometimes functions registered to tracepoints were called even if
they are disabled.
  - So I guarded the calling for disabled tracepoints in 72fff22bca1a93af6b909351994eb13265127fb0. It protects us from SEGV.
- It was hard to remove failure for stopping Byebug in the case that interface input is `nil`.
  - But I want to continue by `C-d` and stopping Byebug. I fixed `LocalInterface` to use `continue` when C-d (EOF) is input. It is just a workaround but works fine. faa5894455f5f175a2562496f0e8a32b258a98aa
- If Byebug is stopped, you can't get `Byebug.current_context`.
  - I stubbed `Byebug.stop` if there are no tracepoints or catchpoints. fc1eb51ef90b5a6238fe254d323cf34b3ef03b68
     - I added an extra test to ensure `continue` works without stub and it stops Byebug.

@deivid-rodriguez I managed to create working patch and make all tests passing. Are you OK to merge and release this?

deivid-rodriguez:
Thanks! I'll review this very soon!!

k0kubun:
I confirmed that the first implementation works well, but it seems that current implementation locks threads except current threads. With commit 72fff22bca1a93af6b909351994eb13265127fb0, I prevented to release thread locks.

I tried delayed stopping in 2a77361cf1ff812ee50086016220493b9265c202 (another branch) and released locks in `cleanup()`. But it looked like some threads were still locked in WEBrick.

@deivid-rodriguez Do you have any idea to release locks of all threads?

k0kubun:
I noticed that some threads are waiting before `acquire_lock()` and lock after scheduled stop in 2a77361cf1ff812ee50086016220493b9265c202.
I'll fix this patch to avoid unexpected thread locking.

deivid-rodriguez:
Thanks for your hard work!

k0kubun:
I found a safe way to unlock all threads after `Byebug.stop`. https://github.com/k0kubun/byebug/commit/0696f940b4c3107b8ad475596422fb7d5a03ffad
I confirmed that this patch works with my rails app and it didn't cause SEGV or deadlock.

@deivid-rodriguez Could you review this again?

ext/byebug/byebug.c
```diff
@@ -579,7 +579,6 @@ Stop(VALUE self)
 
     breakpoints = Qnil;
     catchpoints = Qnil;
-    threads = Qnil;
```
k0kubun: `cleanup_dead_threads()` depends on `threads` availability. Because `release_lock()` calls `cleanup_dead_threads()`, immediately clearing `threads` causes nil reference by locked threads in `acquire_lock()`.

By keeping `threads`, we can safely release thread locks in `EVENT_TEARDOWN`. This variable is re-assigned with `create_threads_table()` in Byebug.start. So there is no need to clear `threads` in this place.

deivid-rodriguez:
@k0kubun I've started reviewing this.

I actually like the approach @os97673 took for `debase` in https://github.com/denofevil/debase/commit/2864440e1b5278c6d2dd7c4f4a2c66bca773384c, but since we already have something working here, let's get it in for now.

k0kubun:
> I actually like the approach os97673 took for debase in denofevil/debase@2864440, 

I agree with you. It looks better to me.

> but since we already have something working here, let's get it in for now.

Thank you. :)

For https://github.com/k0kubun/byebug/commit/f2a97e82b4f9c07a9fde8c5352cea4dea83e417a#commitcomment-13831189, I dropped stub of `Byebug.stop` and commited https://github.com/k0kubun/byebug/commit/f8438279e30094119199caaa62c4b9e0b3765df1. After researching failing tests with the situation, I've noticed that following two checks are necessary too. Thank you for your review.

- post_mortem mode
  - And I removed `ensure`
- current_context's steps
  - for `interrupt`

So I created `Byebug.stoppable?` in https://github.com/k0kubun/byebug/commit/51c4239f46dea4d35c759e40f0636e8c5e3b6542.
While travis CI for OSX seems stopping now, this patch passes AppVayor CI and I confirmed that tests are passing on my OSX and Linux machines.

deivid-rodriguez:
@k0kubun Bad news. I downloaded the patch and run the rake task `loop_tests` (which just run the tests in a loop against the different rubies supported). I found out that the tests get aborted sometimes, every 4 or 5 runs in my case. :( Can you reproduce this?

k0kubun:
They never fail in my Linux environment but I could reproduce random failure in my OSX environment.

It seems that https://github.com/k0kubun/byebug/blob/51c4239/test/commands/break_test.rb#L200-L205 is unstable. I'm researching the reason.

deivid-rodriguez:
Great, thanks!

deivid-rodriguez:
@k0kubun While we sort out the latest problems in this PR, I've cherry-picked into master a couple of commits that are actually independent. So you can remove them from the PR.

k0kubun:
> While we sort out the latest problems in this PR, I've cherry-picked into master a couple of commits that are actually independent. So you can remove them from the PR.

Thank you for cherry-picking :)
 I've just rebased this branch against master and force-pushed.

I continue to investigate the problem.

deivid-rodriguez:
@k0kubun Just so you know, the current test failure in Travis is unrelated to this PR, since I've seen it happening without it. However, the other problem you are investigating does seem related.

k0kubun:
Yes, I'm investigating a random SEGV, which didn't happen on the last CI result.

deivid-rodriguez:
@k0kubun If you want to make sure you get attention from ruby-core, you might want to open an issue at https://bugs.ruby-lang.org and attach you patch there (see https://bugs.ruby-lang.org/projects/ruby/wiki/HowToContribute).

k0kubun:
Thank you. I opened an issue and attached a patch in https://bugs.ruby-lang.org/issues/11603.

deivid-rodriguez:
@k0kubun Just to keep the loop opened here. Does your patch to ruby-core fix the issue with the tests being aborted or does it fix a different issue?

k0kubun:
My patch does fix the issue with the tests being aborted. I've never seen SEGV with the patch.

deivid-rodriguez:
Yes, it does seem to fix it! I just run the tests 50 times in a row using your patch and they never segfaulted, Thanks a lot for this!

Now I'm not sure what's the best path to follow. If I merge we get the speed up in the master branch but we get an unstable build. If I wait, the build doesn't get polluted but we have to run byebug against your branch to get the speed up but 

Do you mind waiting for this to be merged?

k0kubun:
> Do you mind waiting for this to be merged?

We are alreadly using this patch in our project with my fork and have nothing to be worried.

Feel free to leave this open to keep build stable. :)

deivid-rodriguez:
:+1: I'll start using this branch too! :)

brian-vogogo:
Does it make sense for others to try this branch? 

No improvement here. Page load time still increased from 5s to 18s after using binding.pry. (Added binding.pry, viewed page, exit pry, removed binding.pry, reloaded page).

Gemfile changes:
```ruby
  gem 'pry', '0.10.3',
      :git => 'https://github.com/pry/pry',
      :branch => '0-10-stable'
  gem 'byebug', '6.0.2',
    :git => 'https://github.com/k0kubun/byebug.git',
    :branch => 'enable-disable-tracepoints'
  gem 'pry-byebug', '3.2.0',
      :git => 'https://github.com/deivid-rodriguez/pry-byebug'
```
We are on ruby 2.1.6, rails 4.1.10. Do not have the ruby/ruby#1059 patch but assume that is just for stability.

k0kubun:
This branch just fixes `byebug` or `debugger`. You should use https://github.com/deivid-rodriguez/pry-byebug/pull/80 too for `binding.pry`.

```rb
gem 'pry-byebug', git: 'https://github.com/k0kubun/pry-byebug', branch: 'byebug-stop'
```

deivid-rodriguez:
I think you would need to use https://github.com/deivid-rodriguez/pry-byebug/pull/80 as well if you want to try it with `pry-byebug`.

brian-vogogo:
Still no luck. I tried a fresh rails app: https://github.com/brian-vogogo/blog. What version of ruby/rails are you using? Feel free to ignore if you think this might be quirk of my environment. Thanks!

k0kubun:
I tried your app and it worked fine. Did you try `spring stop` if you are using spring?

brian-vogogo:
No spring (commented out gem, checked with `ps aux). Updated sample app with latest ruby (2.2.3) and latest rails (4.2.4). No improvement (25ms vs 400ms).
```
Completed 200 OK in 23ms (Views: 22.9ms | ActiveRecord: 0.0ms)
Completed 200 OK in 27ms (Views: 26.6ms | ActiveRecord: 0.0ms)

# added binding.pry

From: /Users/brian/projects/faster-pry/blog/app/controllers/welcome_controller.rb @ line 4 WelcomeController#index:

    2: def index
    3: 	binding.pry
 => 4: 	puts 'hello'
    5: end

[1] pry(#<WelcomeController>)> exit
hello
  Rendered welcome/index.html.erb within layouts/application (0.9ms)
Completed 200 OK in 3312ms (Views: 414.4ms | ActiveRecord: 0.0ms)

# commented out binding.pry

Completed 200 OK in 402ms (Views: 395.1ms | ActiveRecord: 0.0ms)
Completed 200 OK in 488ms (Views: 478.7ms | ActiveRecord: 0.0ms)
Completed 200 OK in 386ms (Views: 381.4ms | ActiveRecord: 0.0ms)
```

k0kubun:
> No spring (commented out gem, checked with `ps aux). Updated sample app with latest ruby (2.2.3) and latest rails (4.2.4). No improvement (25ms vs 400ms).

Hmm.. I can't think of any possibility to reproduce the situation. But I'm interested in the problem because my colleague also said the same thing.

Again, could you check the return value of `Byebug.started?`? You can check it by

```rb
  def index
  	binding.pry
  	puts "hello: #{Byebug.started?}"
  end
```

or something. And there are two ways to exit the REPL of pry-byebug. Which way do you use, Control-D or `continue`? (Though I supported both situations in https://github.com/deivid-rodriguez/pry-byebug/pull/80.)

k0kubun:
I got it. Your log said you exit REPL with `exit`. I'll fix the patch... thanks.

k0kubun:
@brian-vogogo I optimized pry-byebug for `exit` too in https://github.com/deivid-rodriguez/pry-byebug/commit/de89e05b6ec7c292e294754e3b293fbaa51dcb22. Could you try the new revision of byebug-stop branch?

brian-vogogo:
Thanks!
`exit` is fast
`quit` is fast
`Ctrl-D` is fast
`continue` is slow

k0kubun:
I'm glad to hear that. :)

> continue is slow

Indeed, it does. I overlooked :breakout_nav is thrown before `Byebug.stop`... I fixed it. I think all of them are fast now.

brian-vogogo:
All are fast. Thanks so much for working on these PRs!

deivid-rodriguez:
> All are fast. Thanks so much for working on these PRs!

Yes, thanks! :green_heart:

deivid-rodriguez:
@k0kubun Would you mind rebasing this on top of current master so we can use the latest fixes I've added to master as well?

k0kubun:
Done.

deivid-rodriguez:
Thanks!

deivid-rodriguez:
@k0kubun I've changed the travis build to use your patch to Ruby so this should be ok to merge. I'm not sure whether the segfault could affect normal usage but since you've been using this with success for a while, I'm going to merge it.

Can you rebase one last time and add a CHANGELOG entry?

k0kubun:
> I've changed the travis build to use your patch to Ruby so this should be ok to merge. I'm not sure whether the segfault could affect normal usage but since you've been using this with success for a while, I'm going to merge it.
 
Thanks! It's very helpful for us. We didn't see SEGV on our normal usage.

Rebased and added a CHANGELOG entry.

deivid-rodriguez:
Finally merged this, the appveyor build now segfaults sometimes but I'm happy to have both an stable travis build and a fast debugging feature on master!

I really appreciate the work you did here. Thanks a lot @k0kubun !! :heart: :green_heart: :yellow_heart:

k0kubun:
It was interesting to investigate this problem for me and I'm very happy to contribute to this awesome project!
Thank you for your helpful comments too. :)

deivid-rodriguez:
:clap: :clap: :clap:

PR #167, opened 2015-08-27
Don't load entire library up front

### LINKED ISSUE #166: Lazy loading of Byebug
I’m working on a project where we like to always have Byebug available in our test suite. But calling `require 'byebug'` in our spec helper adds 160ms to our RSpec startup time. And we want that number to be as low as possible.

Therefore, in our spec helper, I have defined this little hack:
```rb
def byebug
  require 'byebug'
  super
end
```

I’m wondering if we could implement something similar in Byebug itself, so I could just call something like `require 'byebug/lazy'` instead of the hack above?

bquorning:
When running tests, it is nice to both have the `byebug` command available, and to have the tests start up really fast.

Fixes #166.

### DIFF
--- a/lib/byebug.rb
+++ b/lib/byebug.rb
@@ -1,2 +1,1 @@
-require 'byebug/core'
 require 'byebug/attacher'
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks!

Issue #47, opened 2014-01-15
Break on  application-code only


sr3d:
Hi,

Byebug is great and it seems to be a lot less buggy than debugger so thumbs up!   However, the issue I'm running into is that whenever I step in, byebug (and also debugger) would step in too deep, or into a chunk of code that's somewhere deep in the gem.  Obviously I don't want to step thru any gem code or external library but focus only on my current application's code.

My question/suggestion is that if there's a way to filter or skip over external library, or a setting to step into the next line that belongs to a script within the current application.  For example, step into a method call will skip over any Rails's internal script or any currently-used gem and stop at the next line of the file that's inside the application.  This will greatly reduce the tediousness of stepping into the app and help pin-pointing the issue.  Currently it's quite time consuming to setup the correct breakpoint, and usually it's a hit-or-miss and having to run the test suite multiple times just to figure out the correct breakpoint placement.

Something like a configuration to list the path or locations to resume the debugger when stepping in automatically would be a great start.

deivid-rodriguez:
Changed my mind again after rereading this issue and [this stackoverflow question](http://stackoverflow.com/questions/28254262/byebug-gem-configuraiton-issues).

bnolan:
:+1:  I'd like to be able to do this.

deivid-rodriguez:
As a tip to anyone wanting to contribute here, debase just added file filtering: https://github.com/denofevil/debase/commit/722157903d50a165371fc988a96c0f96424a17b0

sundarv85:
It would be great to have file filter where one can specific the list of files or the directories in the path where it should step through.

danielyaa5:
🏏 🏏 🏏

nguyenquangminh0711:
I think this could solve your issue: https://rubyjard.org/. It's a wrapper around byebug, with a nice Terminal UI, and plenty of features. One of the key feature is frame filtering that allow you to debug application code only, or include/exclude any gems/ruby libs you want.

lazyatom:
I would really love this too. Rubyjard looks excellent but when not in a terminal (i.e. running a debugger within an editor), it's no longer suitable.

From what I can tell, Rubyjard handles this by wrapping the command processor and stepping through until the current file matches a set that it should stop on, or some execution time or depth expires. Is that a suitable solution here, or does it need to be implemented somewhere in the C bindings, like debaser has done?

nguyenquangminh0711:
[...]

PR #169, opened 2015-08-31
Listing if there's no source file

raffapen:
A common scenario of missing source files is debugging through dynamically generated code. In this case (and probably in any other case where source file is not to be found), it may be better to step through the source-less code and to resume debugging at the next available source line.
Anyway, just failing on missing source file crashes byebug on dynamically generated code.

### DIFF
--- a/lib/byebug/commands/list.rb
+++ b/lib/byebug/commands/list.rb
@@ -35,8 +35,14 @@ module Byebug
     end
 
     def execute
-      msg = "No sourcefile available for #{frame.file}"
-      fail(msg) unless File.exist?(frame.file)
+      if !File.exist?(frame.file)
+        context.step_into 1
+        processor.proceed!
+        return
+      end
+
+      # msg = "No sourcefile available for #{frame.file}"
+      # fail(msg) unless File.exist?(frame.file)
 
       max_lines = n_lines(frame.file)
       b, e = range(@match[2], max_lines)


deivid-rodriguez:
Hi, thanks!

I'm not sure this issue happens on last master, could you add a test so I can understand what we are fixing here? I think source location is properly reported for dinamically generated methods.

deivid-rodriguez:
Hi, thanks!

I'm not sure this issue happens on last master, could you add a test so I can understand what we are fixing here? I think source location is properly reported for dinamically generated methods.

os97673:
@deivid-rodriguez sometimes file of the event could be nil (e.g. if you create and use enumerator) perhaps you throw NoSuchMethod somewhere in https://github.com/deivid-rodriguez/byebug/blob/master/lib/byebug/helpers/file.rb (we had similar problem in ruby-debug-ide)

deivid-rodriguez:
I'd like to know what exact issue the OP is having. That check is meant for `-e` contexts, not for dinamically generated method contexts, and it should never crash `byebug`.

In any case, I don't like the solution since I would never expect a list command to actually step through code...

(Reopening, by the way, I didn't mean to close and comment, just comment).

raffapen:
OP here.
Please consider the following: https://gist.github.com/3cbd4d10cb5f051e9d90.git
Environment:
- Running on Windows (!) with ruby 2.0.0p451 (x64-mingw32),

With byebug 3.5.1, 5.0, experience of stepping into constructor:
(byebug) s
*** No sourcefile available for (eval)
(byebug) s
*** No sourcefile available for (eval)
(byebug) s
*** No sourcefile available for (eval)

(which is normal, but a bit unpleasant.)

With byebug 6.0.2:
C:/Gems/gems/byebug-6.0.2/lib/byebug/commands/list.rb:39:in `execute': No sourcefile available for (eval) (RuntimeError)
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:119:in `block in run_auto_commands'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:119:in `each'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:119:in `run_auto_commands'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:126:in `before_repl'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:93:in `process_commands'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/processors/command_processor.rb:75:in `at_line'
        from C:/Gems/gems/byebug-6.0.2/lib/byebug/context.rb:109:in `at_line'
        from (eval):2:in `A'
        from class1.rb:145:in `<main>'

I guess that under these circumstances, the better approach will be letting 'step' continue stepping until a viable source line is found (rather than let 'list' do the same). However, if one miraculously land on invisible source, 'list' might as well take an active approach.

Thank you.

deivid-rodriguez:
Thanks @raffapen. I can't access your gist though. Also, are you able to try this with one of the versions recommended in the README?

deivid-rodriguez:
Actually the recommended versions are not yet available in Windows so 2.0 is fine.

raffapen:
Fixed gist: https://gist.github.com/raffapen/3cbd4d10cb5f051e9d90

deivid-rodriguez:
Ok, I downloaded the file. What do I have to run to trigger the error?

deivid-rodriguez:
Got it! It was easy :) Let me have a look at it!

raffapen:
Thanks!
:)

deivid-rodriguez:
Wow, there's some heavy metaprogramming in there. My observations:

- This should definitely be fixed in `byebug`. Need to think of the best way though.
- In the mean time, adding `__FILE__` and `__LINE__` parameters to your `class_eval` calls should fix the error. Instead of `class_eval("@@ctors ||= []")` do `class_eval("@@ctors ||= []", __FILE__, __LINE__)` and when using a heredoc, pass `__LINE__ + 1` instead of `__LINE__`.

raffapen:
Great! I'll give it a try.
Thanks for the quick workaround!
BTW, I might not be a bad idea to have a pragma of some sort that tells byebug to avoid stepping through certain places in code. It may serve metaprogramming endeavors. :)

deivid-rodriguez:
Yep, there's a feature request for that already: https://github.com/deivid-rodriguez/byebug/issues/47

deivid-rodriguez:
Just fixed the issue by restoring the old behavior.

I think it's better to encourage the user to wonder why the source file is not available. Maybe they'll find out about the `__FILE__`, `__LINE__` arguments, about the block version of `class_eval` or maybe they realize that their code is abusing metaprogramming. ;)

Cheers!

PR #180, opened 2015-10-31
Fixing setting line trace name on example

guilherme:
It was calling: 'set trace' instead of 'set linetrace'.


deivid-rodriguez:
Thanks!

douglasresende:
:+1:

PR #186, opened 2015-11-10
Adding missing require to processor files

guilherme:
As We discussed in: https://github.com/deivid-rodriguez/byebug/commit/da69f65a, adding some missing requires.

### DIFF
--- a/lib/byebug/processors/control_processor.rb
+++ b/lib/byebug/processors/control_processor.rb
@@ -1,3 +1,5 @@
+require 'byebug/processors/command_processor'
+
 module Byebug
   #
   # Processes commands when there's not program running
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Thanks!

I know I'm very pick with style but, could you add a blank line after the require? I always leave a blank line between requires and the beginning of the code.

guilherme:
Yeah, that's ok :+1: i've added it on: https://github.com/guilherme/byebug/commit/e425958d122d07ee9b5f4071dd382615c62a0433 

I'm not sure if it was the right approach, if you will squash it or i could do it here.

deivid-rodriguez:
Yes, could you squash them?

deivid-rodriguez:
Let me know If you need help with that.

guilherme:
done :+1:

PR #190, opened 2015-11-19
Return true as boolean if byebug is already started

Sheile:
This is very minor issue.

`Byebug.started?` return `false` when breakpoints hasn't been set.
`Byebug.started?` return `0` when some breakpoints has been set.


deivid-rodriguez:
Doesn't it always return `0` or `1`? In any case, yes, it's supposed to return a Ruby boolean, so merging.

PR #206, opened 2016-02-01
Added an example to the usage

jamtur01:
Expanded the Usage example to make the workflow a bit clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,22 @@ Simply drop
 
     byebug
 
-wherever you want to start debugging and the execution will stop there. If you
-are debugging rails, start the server and once the execution gets to your
-`byebug` command you will get a debugging prompt.
+wherever you want to start debugging and the execution will stop there.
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hi @jamtur01, thanks for the PR! The example is now much easier to understand!

Just some style nitpicks:

* I recently added a markdown linter to keep a consistent style in the markdown files too. Could you check the build and fix the style issues?
* Could you use `shell` instead of `Shell` and `ruby` instead of `Ruby` in the code blocks?
* Could you use `bin/rails` instead of `rails` in the example?

Thanks!

jamtur01:
Think that should fix it up. Can squash if you'd prefer too.

deivid-rodriguez:
Yes, please! :)

jamtur01:
Done!

deivid-rodriguez:
Thanks!

Issue #197, opened 2015-12-17
Can't launch irb in ruby-2.0.0-p647 because irb parses ARGV


josephks:
When I try to invoke irb from within byebug inside my cucumber scenario I can't launch irb because irb tries to parse ARGV which contains cucumber's command line options.  This wasn't a problem for me in ruby 1.9.3 with debugger.

Modifying the execute method in lib/byebug/commands/irb.rb like this solves my problem:

```
    def execute
      unless processor.interface.is_a?(LocalInterface)
        return errmsg(pr('base.errors.only_local'))
      end
      saved_argv = ARGV.dup
      ARGV.clear
      IRB.start(__FILE__)
      ARGV.concat(saved_argv)
    end
```

deivid-rodriguez:
Could you post some steps to reproduce this?

josephks:
* Put `debugger` in your code
* Run `cucumber -b -r features`
* When you enter the debugger type `irb`
* Watch irb blow up because it can't partse "-b -r features'

deivid-rodriguez:
Can you post a sample app? I don't want to deal with installing and setting up cucumber. You can also create a PR with a test case.

deivid-rodriguez:
Thanks @josephks, I think you patch makes perfect sense. Could you create a PR?

PR #208, opened 2016-02-02
Fix issue 197 by clearing and restoring ARGV

josephks:
See #197 

Test results are the same on this branch as on master.

I tested manually with https://github.com/josephks/bug_demo and this fix works.  I'm not sure how to write an automated test for this.

### DIFF
--- a/lib/byebug/commands/irb.rb
+++ b/lib/byebug/commands/irb.rb
@@ -29,7 +29,20 @@ module Byebug
         return errmsg(pr('base.errors.only_local'))
       end
 
-      IRB.start(__FILE__)
+      # IRB tries to parse ARGV so we must clear it.  See issue 197
+      with_clean_argv { IRB.start(__FILE__) }
+    end
+
+    private
+
+    def with_clean_argv
+      saved_argv = ARGV.dup
+      ARGV.clear
+      begin
+        yield
+      ensure
+        ARGV.concat(saved_argv)
+      end
     end
   end
 end


deivid-rodriguez:
Yep, it's not easy to write an automated test for this. I think we can live without it!

Can you change the code to look like this:

```ruby
with_clean_argv { IRB.start(__FILE__) }
```

and move all the ARGV handling logic to a separate `with_clean_argv` private method?

josephks:
Updated

deivid-rodriguez:
@josephks Just one more nitpick and I'm happy to merge. Now that this is extracted into a method, you no longer need the explicit `begin`, the beginning of the method does the same job. Could you remove it?

Thanks! :)

josephks:
:question: I don't have a begin at the beginning of the method.

deivid-rodriguez:
Oh sorry, I mean like this

```ruby
def with_clean_argv
  saved_argv = ARGV.dup
  ARGV.clear
  yield
ensure
  ARGV.concat(saved_argv)
end
```

josephks:
I didn't do that on purpose.  I only want to restore ARGV after it has been cleared.  If somehow something went wrong with the dup or the clear then I don't want to restore.

deivid-rodriguez:
I think if `ARGV.dup` or `ARGV.clear` go wrong... the user has bigger problems with her system than this one... :)

Anyways, squash to a single commit and I'll merge and release a new version with the fix. Thanks for your contribution @josephks !!

josephks:
Squashed

PR #210, opened 2016-02-05
Better file path acceptance for b[reak]

HookyQR:
Enables adding break points with files names such as:
```
../Test/Relative/path.rb
/Test Space/path.rb
c:/Test_Windows/path.rb
```

Also eliminates crash when forgetting to type the line number (oops :wink:) with a relative path. Currently: `break ../file.rb` will crash as it attempts to resolve that line as a `class.method` of `/file.rb`.

### DIFF
--- a/lib/byebug/commands/break.rb
+++ b/lib/byebug/commands/break.rb
@@ -15,7 +15,7 @@ module Byebug
     self.allow_in_control = true
 
     def self.regexp
-      /^\s* b(?:reak)? (?:\s+ (.+))? (?:\s+ if \s+(.+))? \s*$/x
+      /^\s* b(?:reak)? (?:\s+ (.+?))? (?:\s+ if \s+(.+))? \s*$/x
     end
 
     def self.description
--- a/lib/byebug/commands/break.rb
+++ b/lib/byebug/commands/break.rb
@@ -15,7 +15,7 @@ module Byebug
     self.allow_in_control = true
 
     def self.regexp
-      /^\s* b(?:reak)? (?:\s+ (\S+))? (?:\s+ if \s+(.+))? \s*$/x
+      /^\s* b(?:reak)? (?:\s+ (.+))? (?:\s+ if \s+(.+))? \s*$/x
     end
 
     def self.description
@@ -52,7 +52,7 @@ module Byebug
 
     def line_breakpoint(location)
       line_match = location.match(/^(\d+)$/)
-      file_line_match = location.match(/^([^:]+):(\d+)$/)
+      file_line_match = location.match(/^(.+):(\d+)$/)
       return unless line_match || file_line_match
 
       file = line_match ? frame.file : file_line_match[1]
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Hi @HookyQR, thanks for this. The build is failing though, could you have a look?

HookyQR:
appveyor has been failing for quite a few release. The failure is on the same errors as the Ruby 2.0 runs on travis-ci.

travis-ci is failing for 2.1.8 on an install error with `gem install ffi -v '1.9.10'`. The rest of the (not fail-able) tests are now passing (removed greedy match). Note that I have made the same relative change for the `info file` command.

deivid-rodriguez:
@HookyQR, right. Appveyor should get fixed once `2.1.8`, `2.2.4` and `2.3.0` land in there. Should happen soon. Regarding travis & 2.1.8, not your fault either, looks like a travis issue: https://github.com/travis-ci/travis-ci/issues/5580.

So it looks like your patch does not break anything. I would like some tests proving the functionality you're adding though. Basically, what you described in your first comment.

HookyQR:
Tests added. Used files from the test suite to check relative matching.

deivid-rodriguez:
@HookyQR Thanks. I have some questions though:

I added the following two tests on top of current master and they're already passing.

```diff
diff --git a/test/commands/break_test.rb b/test/commands/break_test.rb
index 626a1d3..0c81c2f 100644
--- a/test/commands/break_test.rb
+++ b/test/commands/break_test.rb
@@ -169,6 +169,20 @@ module Byebug
       check_error_includes 'No file named asf'
     end
 
+    def test_setting_breakpoint_with_bad_relative_path_doesnt_crash
+      enter 'break ../relative/path.rb:8'
+      debug_code(program)
+
+      check_error_includes 'No file named ../relative/path.rb'
+    end
+
+    def test_setting_breakpoint_with_relative_path
+      enter 'break ./test/commands/break_test.rb:8'
+      debug_code(program)
+
+      check_output_includes(/Successfully created breakpoint with id/)
+    end
+
     def test_setting_breakpoint_to_invalid_line_does_not_create_breakpoint
       enter 'break 14'
``` 

So it seems to me that this PR is not about handling relative paths (which already works) but handling paths with spaces. Is that correct?

HookyQR:
Reviewing the code, I guess it is just about the spaces in filenames/paths. I think I had made some change along the way that broke the relative path matching and probably reverted that before the commit. Sorry for the misinformation.

deivid-rodriguez:
@HookyQR No problem! Now the PR should be simpler to review.

Ideally, I'd like you to rebase against latest master (I've added the tests I mentioned) and squash this PR into 2 commits, "Allow for space in path in 'info file'" and "Allow for space in path in 'break'", each containing the changes for each command and the respective tests. Could you do that?

deivid-rodriguez:
@HookyQR Both Travis and Appveyor seem fixed now, so you should get a green build after reworking git history and force pushing!

PR #211, opened 2016-02-10
Use IO.console for readline over STD{IN,OUT,ERR}.

sethk:
I've been testing this for a few weeks to debug programs that read from stdin and write to stdout. This is the way I'm accustomed to other debuggers working out of the box.

### DIFF
--- a/lib/byebug/interfaces/local_interface.rb
+++ b/lib/byebug/interfaces/local_interface.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'io/console'
+
 module Byebug
   #
   # Interface class for standard byebug use.
@@ -8,9 +10,9 @@ module Byebug
 
     def initialize
       super()
-      @input = STDIN
-      @output = STDOUT
-      @error = STDERR
+      @input = @output = @error = IO.console
+      Readline.input = @input
+      Readline.output = @output
     end
 
     #


deivid-rodriguez:
@sethk Thanks. Could you teach me which problem this solves? I've never used `IO.console` before...

sethk:
`IO.console` is a wrapper for `/dev/tty`, which is the controlling terminal for the current session (usually the shell). This means you can interact with the user even if stdin/stdout are being redirected.

Let's say you have a small script like `rev.rb`:
```
#!/usr/bin/env ruby -w

ARGF.each { |line| print line.reverse }
```

Run it against the system dictionary:
```
ruby rev.rb < /usr/share/dict/words
```

But if you try to debug with `byebug`, it tries to interpret the input as debugger commands:
```
byebug rev.rb < /usr/share/dict/words
```

But with my patch it works correctly, and there is no change unless `STDIN.tty? || STDOUT.tty?` If you want, I can special case it to only use `IO.console` in that case.

deivid-rodriguez:
Sweet! This is great!! :)

Just 2 comments:

- Can you think of a way to test this?
- Can you move the require under the magic comment? Otherwise it won't be picked up and frozen string literals won't be enforced.

sethk:
As far as testing goes, it's possible to use a PTY (pseudo terminal) to simulate input for Readline. If you spawned a Byebug session connected to a PTY, with the PTY's FD used for debugger I/O, and STD{IN,OUT} used by the Ruby script itself, it might prevent you from having to use Byebug::TestInterface as much and make your tests closer to reality. On the other hand, it may not be worth the trouble, as the many different Readline versions should have their own tests for interacting directly with the terminal.

sethk:
Fixed the magic string and squashed the commits.

deivid-rodriguez:
Thank you @sethk !

PR #219, opened 2016-03-16
Update Guide.md to use 'eval' instead of 'p'

hlindberg:
The `eval` command works better for printing out variables than `p` since `p` will first print, and then the result of `p` is again printed.
The result is that all output gets displayed twice.

This simply changes the `p` to `eval` in the guide.


deivid-rodriguez:
Great! Thanks!

PR #229, opened 2016-04-06
Log .byebugrc errors to the console

sethk:
I notice there used to be a verbose option, which I think could be used to echo ``.byebugrc`` commands as they are executed. For now the verbose option of ``ScriptInterface`` doesn't seem to be connected to anything.

### DIFF
--- a/lib/byebug/interfaces/script_interface.rb
+++ b/lib/byebug/interfaces/script_interface.rb
@@ -1,3 +1,5 @@
+require 'io/console'
+
 module Byebug
   #
   # Interface class for command execution from script files.
@@ -6,8 +8,9 @@ module Byebug
     def initialize(file, verbose = false)
       super()
       @input = File.open(file)
-      @output = verbose ? STDOUT : StringIO.new
-      @error = verbose ? STDERR : StringIO.new
+      @verbose = verbose
+      @error = IO.console
+      @output = verbose ? @error : File.open(File::NULL, File::WRONLY)
     end
 
     def read_command(prompt)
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks @sethk!

So before this, no output would be printed at all if we put, say, an invalid command inside `.byebugrc`? If that's the case, can we add a test for that?

sethk:
Because of how `/dev/tty` works, I don't think there's a straightforward way to test for this. Something I thought about is passing `input`/`output`/`error` to the `Interface` initializer and nixing the `TestInterface` class. That way you could pass `StringIO`s to any of the interfaces for testing.

deivid-rodriguez:
>  Something I thought about is passing input/output/error to the Interface initializer and nixing the TestInterface class. That way you could pass StringIOs to any of the interfaces for testing.

This is a very good idea. I wanted to do it for a long time but never got to it. Do you think you can get the ball rolling? Maybe not completely ditch the `TestInterface` class for now if it's too much work but figure this better dependency injection mechanism out so we can start testing these kind of things.

deivid-rodriguez:
@sethk I did this myself in #267 without `IO.console`. We can think of adding that later. Can you have a look.

Closing in favor of #267

deivid-rodriguez:
@sethk Just released 9.0.4 including this. Let me know!

sethk:
Sorry for going AWOL. I will look into this and share some thoughts about refactoring soon.

sethk:
Okay, now I can at least see the errors my startup script is generating. =)

PR #238, opened 2016-04-18
Fix creation of remote control processor

izaera:
Please consider merging this fix for remote debugging.

Cheers,
Ivan.


deivid-rodriguez:
@izaera Totally up for keeping more explicit dependencies, specially when not doing it leads to errors!

Just one tiny change, can you use regular `require` instead of `require_relative`, so we stay consistent with the rest of the code?

izaera:
OK. I’ll redo and resend.

Thx.

PR #250, opened 2016-05-06
Use global instead of constant stdin, stdout and stderr

bouk:
This makes it possible to use http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-capture_io to capture the output of byebug in tests (more context: https://github.com/deivid-rodriguez/pry-byebug/pull/90)

### DIFF
--- a/lib/byebug/interfaces/local_interface.rb
+++ b/lib/byebug/interfaces/local_interface.rb
@@ -8,9 +8,9 @@ module Byebug
 
     def initialize
       super()
-      @input = STDIN
-      @output = STDOUT
-      @error = STDERR
+      @input = $stdin
+      @output = $stdout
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bouk It seems that you managed to solve this issue with testing pry-byebug through other mechanism, so since this part of the code is very likely to change in the near future (to fix #229), I'll close this.

I'll get to your PR in `pry-byebug` soon.

Sounds good?

deivid-rodriguez:
@bouk I'm doing some work for #229 and I think this is a good easy way to get it tested. Could you rebase this so I can merge it?

deivid-rodriguez:
Thank you! :)

PR #262, opened 2016-05-14
Fix build failure on "ruby-head" x "CC=clang"

yui-knk:
"clang: error: unknown argument: '-fexcess-precision=standard'"
occurs when "compiling ../../../../ext/byebug/breakpoint.c".

The reason why compile error occurs is that we use ruby build
by gcc and compile byebug with clang.
"mkmf" uses values of `RbConfig` module which were defined
when ruby is built.

This commit make "env CC" match build information of ruby.

ref: https://github.com/ruby/ruby/blob/v2_3_0/tool/mkconfig.rb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,17 +22,18 @@ script:
 
 rvm:
   - 2.1.10
+  - 2.1.10-clang
   - 2.2.5
+  - 2.2.5-clang
   - 2.3.1
+  - 2.3.1-clang
   - ruby-head
-
-env:
-  - CC=gcc
-  - CC=clang
+  - ruby-head-clang
 
 matrix:
   allow_failures:
     - rvm: ruby-head
+    - rvm: ruby-head-clang
 
 branches:
   only:


yui-knk:
Or maybe we can remove

```
env:
  - CC=gcc
  - CC=clang
```

and add these versions to `rvm`

* 2.1.10-clang
* 2.2.5-clang
* 2.3.1-clang
* ruby-head-clang

deivid-rodriguez:
Yes, that looks like a good idea!

yui-knk:
Which do you think is better

* Do not remove `env` (now)
* Remove `env`

deivid-rodriguez:
I think if we use the clang rubies provided by travis we don't need to use `env`. So I'd say remove `env`.

deivid-rodriguez:
@yui-knk Looks great now! Can you rebase so I can merge? :)

yui-knk:
@deivid-rodriguez rebased!

deivid-rodriguez:
Thanks a lot!

PR #263, opened 2016-05-14
Skip to get a line in eval context

k0kubun:
Fix for test 2 and 3 in https://github.com/deivid-rodriguez/byebug/issues/194.

We can't open a file named "(eval)". So byebug should skip to read its line on linetrace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Master (Unreleased)
 
+### Fixed
+
+* Skip to get a line in eval context (#263).
+
 ## 9.0.1 - 2016-05-14
 
 * `quit` never exiting when remote debugging (#201).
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@k0kubun Looks great. Just one nitpick, can you split the assertion into a different test? Oh, and can you add a change log entry?

deivid-rodriguez:
Mmm, I meant to split the assertion to a separate method, not a separate class and file. I'm aiming at having one file in `test/commands` per command in `lib/byebug/commands`, that's why I prefer not to have this in a separate file and keep it in the `test/commands/continue_test.rb` file. 

Also, the change log entry is missing a "### Fixed" entry and we need a rebase. 

Sorry for the trouble...

deivid-rodriguez:
Thank you very much @k0kubun!!

k0kubun:
:smile:

PR #264, opened 2016-05-14
Don't stop byebug with linetrace enabled

k0kubun:
Fixes test 1 of https://github.com/deivid-rodriguez/byebug/issues/194.

We should keep tracepoints enabled to trace lines.


deivid-rodriguez:
Thanks a lot! :heart:

PR #275, opened 2016-06-08
Use http://spdx.org/licenses identifier in gemspec.

robertclancy:
Using an spdx license identifier makes it easier for tools like https://github.com/pivotal/LicenseFinder to manage gem licenses.


deivid-rodriguez:
Thanks!

PR #281, opened 2016-07-05
Delete enable disable messages

rocky:
A warmup set of commits. 

These allow a front-end like [realgud](https://github.com/realgud/realgud-byebug) to track what's going on easier.


test/commands/next_test.rb
```diff
@@ -121,6 +121,7 @@ def program
     end
 
     def test_next_works_as_expected_with_define_method
+      skip('Next is 12 not 20 on Ruby 2.2.3') if RUBY_VERSION == '2.2.3'
```
deivid-rodriguez: Please remove this commit and upgrade your ruby instead.

deivid-rodriguez:
Hi @rocky, thanks for your patches and sorry for taking so long to get here. They are now first in the queue, I hope to review soon.

For now I added a comment. The tests don't all pass in ruby 2.2.3, that's expected. Please upgrade your ruby to 2.2.5 and remove that commit.

rocky:
Would you please make the appropriate change?

deivid-rodriguez:
You mean you are not up for addressing any feedback?

rocky:
I won't be able to get to this for a while.

deivid-rodriguez:
Ok!

lib/byebug/breakpoint.rb
```diff
@@ -31,12 +31,13 @@ def self.add(file, line, expr = nil)
...
     #
-    # Removes a breakpoint
+    # Removes a breakpoint and returns the breakpoint information
     #
     # @param id [integer] breakpoint number
     #
     def self.remove(id)
-      Byebug.breakpoints.reject! { |b| b.id == id }
+      bi = Byebug.breakpoints.find_index { |b| b.id == id }
+      bi ? Byebug.breakpoints.delete_at(bi) : nil
```
rocky: @deivid-rodriguez Line 34 suggests what I was thinking of, that change returns the breakpoint object and this has other information about the breakpoint such as whether it had been enabled or disabled,  if there was a condition attached and so on. 

However it looks like the only use of this, the delete command doesn't make use of that additional information. So sure, the only essential part  is in delete.rb and this doesn't need to be modified.

deivid-rodriguez: Ok, let's remove both the change in the code and in the comment then.

lib/byebug/printers/texts/base.yml
```diff
@@ -16,6 +16,7 @@ break:
     not_changed: "Incorrect expression \"{expr}\", breakpoint not changed"
   confirmations:
     delete_all: "Delete all breakpoints?"
+  deleted: "Delete breakpoint {pos}"
```
deivid-rodriguez: I think you added the new message here, but are still using the hardcoded string in the code.

lib/byebug/helpers/toggle.rb
```diff
@@ -26,14 +24,22 @@ def enable_disable_breakpoints(is_enable, args)
           selected_breakpoints = all_breakpoints.select do |b|
             selected_ids.include?(b.id)
           end
+          return selected_breakpoints
```
deivid-rodriguez: Don't use explicit `return` unless necessary. Also, after this refactoring the `selected_breakpoints` variable is no longer necessary.

lib/byebug/breakpoint.rb
rocky: Well, actually the other thing this does is give the caller status as to whether it found the breakpoint to remove or not. So _some_ change is desirable. 

If you view lib/byebug/breakpoint.rb as an API interface returning the information about what was removed is more useful. However if you are view this as just servicing the delete command as it is written, then assuming you address the status aspect, then sure change it.

lib/byebug/helpers/toggle.rb
rocky: This kind stylistic stuff, just do it however you want. If you get to the stage where you have to work across several languages that work differently, perhaps you'll see things differently. 

I could just as easily imagine without this the comment "Explicit is better than implicit so put in a return".

lib/byebug/printers/texts/base.yml
rocky: Ok. If this should be removed, then do it. Thanks.

lib/byebug/helpers/toggle.rb
deivid-rodriguez: I'm taking this as a confirmation that you no longer want to work on this PR, even when you get time. I'm asking you to change those things yourselve, not asking you for permission to change them.

lib/byebug/printers/texts/base.yml
deivid-rodriguez: This message is fine, I was asking you to use it.

lib/byebug/helpers/toggle.rb
rocky: I don't have time to work on this and don't see in the foreseeable future having the time. These changes were motivated by requests from others for better support for byebug in realgud.

The harder it is to get changes in, the less likely I am to work on this but rather on something that is more satisfying, more to my needs, and is less effort. 

I recall someone once asking Matz about working on a standard test suite using rspec or TDD, and his reply as I recall was: "I could spend my time working on the test spec or working on Ruby. Which is more fun?"

deivid-rodriguez: You took more time replying to my comments than it would have taken you to address them. For example, my first request was to remove a single line of code. Instead, you wrote wrote a comment telling me to do it myself. Certainly not the best way to get changes in.

Anyways, I'll do this myself!

rocky: > You took more time replying to my comments than it would have taken you to address them. 

Actually that's not true, if you count the time to _test_ the code and update git. As you know this code was written a while back in an environment when I had byebug set up. I often don't have  that environment as easily accessible. 

Also, the same could be said of you. The time you spend writing the  critique could have been spend just adjusting the code. But the difference is that you'ld adjust it in a way that was pleasing to you and there wouldn't be back and forth. It's not clear that on the second or third try this would be the case for me. 

So, yes, thanks for taking this on and finishing it in a way you find pleasing.

deivid-rodriguez: Of course, you are the one who wants this merged in, not me! And of course my quality standards apply, not yours.

I'd say this is pretty standard in open-source, maybe not in the projects you're used to contributing I guess.

deivid-rodriguez:
Closing in favor of #291

PR #283, opened 2016-07-13
Add commands eval and eval?

rocky:
These evaluate the current current source line.
The ? indicates to evaluate the expression-y part


deivid-rodriguez:
@rocky This needs work and communication.

Not sure I want this, but if it was to be added, I would need you to (at least):

* Add tests for the feature.
* Cleanup all the debugging staff left in there.
* Rename the command so that it doesn't override the `eval` method.
* Explain why you need this and how will you use it.

rocky:
There are a lot of things that aren't strictly needed but are conveniences. This command allows me to understand the consequence of the running what's coming up without advancing the program counter. In a world where one could go backwards in history and redo commands, this would be less useful.

I find myself often wanting to run the command about to be executed or some part of it like the expression-y part. This saves me the trouble of cutting and pasting some part of the source.

As mentioned with respect to other pull request, I can't devote time in the foreseeable future to work on this. If someone else can, great if not, so be it.

deivid-rodriguez:
Even if this was in a mergeable state, I decided I don't want this. Thanks anyways!

PR #284, opened 2016-07-22
Colorize

rocky:
Add "set highight {dark|light|plain}" to colorize source-code listing.

Require linecache2 from gemcutter.org. And line caching is a good thing anyway.


deivid-rodriguez:
Definitely up for adding colors!

Here I'd need:

* All tests passing (`skip` is cheating).
* Remove ruby version conditionals.
* Remove line caching again. If you want that, please add a different PR with tests showing which enhancements you're adding.

After that we can start discussing the feature. I think I would prefer to just colorize if we can, instead of adding a command. What do you think?

rocky:
In my experience there are cases where you don't want to colorize output. Some examples that come to mind is when the colors look awful as might happen if you are using a dark background for a light background theme, or there is an interface outside the debugger that handle showing source in some other way. 

In my other debuggers where I use pygments, it also allows setting the colors via some sort of color style that people can select. See http://python2-trepan.readthedocs.io/en/latest/commands/set/style.html and also http://python2-trepan.readthedocs.io/en/latest/commands/set/highlight.html

As for working on the code, I will not have the time to do this in the foreseeable future. I you don't understand how the code works, I can try to explain that, but otherwise, this will have to wait for someone interested enough in this to see it done in the way that makes you happy.

deivid-rodriguez:
@rocky Thanks for your input, I'll consider it when I have a look at the code. First time I checked, this PR wouldn't even get the tests passing.

deivid-rodriguez:
Leaving this WIP PR opened in case someone wants to take it from here.

PR #296, opened 2016-10-28
Readme: Add description for kill, list

garmoshka-mo:


deivid-rodriguez:
@garmoshka-mo Sorry for not replying earlier, I forgot.

I like this, but only the part adding the links to the guide. I think the subcommands column is fine as it is now. Coud you keep only that part and, ideally, add links for the rest of the commands?

garmoshka-mo:
@deivid-rodriguez Thank you for reply. Unfortunately, I don't have extra time for adding full list of links and, honestly, do not feel temptation to be involved deeper.

I'm curious why short command description seems irrelevant, but I not insist on giving reason. Why I added descriptions - it is because it was (and still) big mystery for me what some commands mean. Links would partially resolve issue, but in inconvenient way (instead of few words, like in traditional `command -h` - user will have to read full documentation with lots of text). To be honest, it feels like decisions here may be driven more by emotions, than by logic. ("I like this" / "I think previous is fine" - instead of explanation of what's wrong with solution)

deivid-rodriguez:
@garmoshka-mo You're absolutely right, sorry for not explaining. The reason is that I feel it's confusing that a column of a table means two different things depending on the row. If you want to have concise command help, there's the `help` command, and the links you've added already partially solve the issue. Anyways,documentation is currently very bad, best way to get to know byebug is to use the builtin help system. 

Again, sorry for not explaining answer and seeming arbitrary, and for taking this long to have a look at such a simple improvement.

garmoshka-mo:
Thank you for justification. 
Btw, `help` gives a clue about command, didn't knew about it, thank you.

deivid-rodriguez:
Thank you!

Now you would have to rebase on top on current master (this is needed so I can merge the PR, a very strict check enabled in the repo). Also, since you are at it, could you squash the two commits into just one? So we don't add unnecessary churn to git history.

garmoshka-mo:
created new branch: https://github.com/deivid-rodriguez/byebug/pull/303

To squash there is a button in github:  
screen shot 2016-11-19 at 21 37 23

PR #311, opened 2016-12-06
Show that byebug can be invoked from the sh

mjskier:
## Description

Add an example on how to invoke byebug from the command line.
This has the advantage of not requiring the user to modify the code just to invoke byebug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ gem install byebug
 
 ## Usage
 
+### From within the Ruby code
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks! :+1:

mjskier:
No, thank YOU for such an awesome tool

deivid-rodriguez:
Thanks! :)

PR #315, opened 2017-01-06
Fix issue with byebug/core.rb unavailable from class methods

markburns:
## Description

When using byebug from within a class method, I often get an error like this:
```
     NoMethodError:
       undefined method `mode=' for Byebug:Module
     # /Users/markburns/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/byebug-9.0.6/lib/byebug/attacher.rb:13:in `attach'
     # /Users/markburns/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/byebug-9.0.6/lib/byebug/attacher.rb:30:in `byebug'
```

This is because `mode=` is defined in `byebug/core.rb` which is required in `test_helper.rb`.
I think it is important to be as close as possible to the files users will require in running the tests. So I believe this is the correct fix.

However, I am sorry that I do not have the time to work out how to create an isolated test to reproduce the problem (as by definition the test suite has already defined `Byebug.mode=`, hence fixing the issue). 

If you know a good way to test this, other than manually, I'd be happy to have a go at adding one.

I have of course confirmed manually that this fixes the problem.

### DIFF
--- a/lib/byebug.rb
+++ b/lib/byebug.rb
@@ -1,1 +1,2 @@
 require 'byebug/attacher'
+require 'byebug/core'
--- a/test/support/test_case.rb
+++ b/test/support/test_case.rb
@@ -1,6 +1,5 @@
 require 'minitest'
 require 'byebug'
-require 'byebug/core'
 require 'byebug/interfaces/test_interface'
 require 'support/utils'
 


deivid-rodriguez:
Hi @markburns, I'm not sure I understand the issue.

`byebug/core` is certainly required somewhere else since `test_helper.rb` is only used for `byebug`'s tests. Does this happen when you run `byebug`'s tests or somewhere else? Could you post some steps so I can reproduce the issue?

Regards!

markburns:
It's actually quite intermittent, but persistent. I've been noticing it for years and thought I'd finally get round to fixing it.
I only ever notice it when trying to execute byebug from a class method. So I'm not sure of the cause exactly. However, a test to replicate it would need to remove the require from test_helper to ensure it isn't already loaded. 
I'll have a look now and see if I can do a gist to replicate it.

deivid-rodriguez:
:+1: Don't worry about a test, a repro is enough.

It's weird since `byebug/core` is required right before calling `mode=`: https://github.com/deivid-rodriguez/byebug/blob/master/lib/byebug/attacher.rb#L10-L13.

markburns:
OK so looking at the code more closely I can see the reason for your confusion. It looks like this should never happen. Also I can't isolate easily. 
I thought it was quite a simple problem with a simple solution. I can assure you that it definitely does happen, but am now at a complete loss for why and under what scenario it does.
Perhaps it is some combination of `.*rc` file setting and/or `pry`.
I'll close for now, and look further into it next time it happens. As when it does happen it seems to happen every time I re-run the test I'm working on.

deivid-rodriguez:
Yes, please, do it. Thanks for reporting and following up.

PR #346, opened 2017-08-24
Update appveyor with trunk, cleanup

MSP-Greg:
## Description

Update appveyor.yml

1. Added trunk as an allowed failure
2. Shorted PATH
3. Removed fast finish

## Requirements

Before submitting the PR make sure the following are checked:

* [x] Wrote [good commit messages][commits].
* [x] Used the same coding conventions as the rest of the project.
* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] All tests are passing.
* [x] The PR relates to *only* one subject.

You might also be asked to:

* [ ] Add tests.
* [ ] Add a Changelog entry if the new code introduces user-observable changes.

[autoclose]: https://github.com/blog/1506-closing-issues-via-pull-requests
[commits]: http://chris.beams.io/posts/git-commit

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,15 @@
 ---
+init:
+  # clean path up to just what we need
+  - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\Git\cmd;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Windows\system32;C:\Windows
+  # download and install current MinGW trunk build, add swap GDBM & OpenSSL packages
+  - if %ruby_version%==_trunk (
+      appveyor DownloadFile https://dl.bintray.com/msp-greg/ruby_windows/ruby_trunk.7z -FileName C:/ruby_trunk.7z &
+      7z x C:/ruby_trunk.7z -oC:/ruby_trunk & C:/ruby_trunk/trunk_install.cmd)
 
 version: "{build}"
 
 install:
-  - set PATH=C:\Ruby%ruby_version%\bin;%PATH%
   - echo %PATH%
 
   - ruby --version
@@ -35,12 +41,14 @@ environment:
     - ruby_version: 23-x64
     - ruby_version: '24'
     - ruby_version: 24-x64
+    - ruby_version: _trunk
 
-  codeclimate_repo_token: '02530029b1e956220f05076c590b84b9ab078362c9083312eb2ad
-41cab138408'
+  codeclimate_repo_token: '02530029b1e956220f05076c590b84b9ab078362c9083312eb2ad41cab138408'
 
 matrix:
   fast_finish: true
+  allow_failures:
+    - ruby_version: _trunk
 
 branches:
   only:


appveyor.yml
```diff
@@ -1,9 +1,15 @@
 ---
+init:
+  # clean path up to just what we need
+  - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\Git\cmd;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Windows\system32;C:\Windows
```
deivid-rodriguez: Is a specific path necessary for the build to pass? I seems simpler to me to add to the existing path whatever binaries you need.

appveyor.yml
```diff
@@ -35,12 +41,13 @@ environment:
...
     - ruby_version: '24'
     - ruby_version: 24-x64
+    - ruby_version: _trunk
 
-  codeclimate_repo_token: '02530029b1e956220f05076c590b84b9ab078362c9083312eb2ad
-41cab138408'
+  codeclimate_repo_token: '02530029b1e956220f05076c590b84b9ab078362c9083312eb2ad41cab138408'
 
 matrix:
-  fast_finish: true
```
deivid-rodriguez: Why removing fast finish?

deivid-rodriguez:
Thanks! Looks good to me, and appveyor liked it too :)

Just had a couple of questions.

appveyor.yml
deivid-rodriguez: Actually I kind of like having an explicit path. Looks good! :+1:

MSP-Greg:
> Is a specific path necessary for the build to pass?

No

> I seems simpler to me to add to the existing path whatever binaries you need.

There's pluses/minuses for both ways of handling `PATH`.  As an example, `C:\Program Files\Git\usr\bin` and `C:\msysy64\usr\bin` have many common files.  I've got a standard install of Git, and `usr/bin` is not normally added to `PATH`.  It just seems safer this way.  File conflicts would not be fun to troubleshoot remotely.

> Why removing fast finish?

Sorry.  My thought is that this seems to be a fairly stable repo, and there aren't typically multiple commits per day.  It's my understanding that `fast_finish` will stop the build on the 1st failed job.  I have also seen job failures from 'internet' issues.  If you're okay with the build stopping due to a failure on  ruby22, I can add it back in.


**PATH value (added line breaks)**
```
echo %PATH%
C:\Perl\site\bin
C:\Perl\bin
C:\Windows\system32
C:\Windows
C:\Windows\System32\Wbem
C:\Windows\System32\WindowsPowerShell\v1.0\
C:\Program Files\7-Zip
C:\Program Files\Microsoft\Web Platform Installer\
C:\Tools\GitVersion
C:\Tools\PsTools
C:\Program Files\Git LFS
C:\Program Files (x86)\Subversion\bin
C:\Program Files\Microsoft SQL Server\120\Tools\Binn\
C:\Program Files\Microsoft SQL Server\Client SDK\ODBC\110\Tools\Binn\
C:\Program Files (x86)\Microsoft SQL Server\120\Tools\Binn\
C:\Program Files\Microsoft SQL Server\120\DTS\Binn\
C:\Program Files (x86)\Microsoft SQL Server\120\Tools\Binn\ManagementStudio\
C:\Tools\WebDriver
C:\Program Files (x86)\Microsoft SDKs\TypeScript\1.4\
C:\Program Files (x86)\Microsoft Visual Studio 12.0\Common7\IDE\PrivateAssemblies\
C:\Program Files (x86)\Microsoft SDKs\Azure\CLI\wbin
C:\Ruby193\bin
C:\Tools\NUnit\bin
C:\Tools\xUnit
C:\Tools\MSpec
C:\Tools\Coverity\bin
C:\Program Files (x86)\CMake\bin
C:\go\bin
C:\Program Files\Java\jdk1.8.0\bin
C:\Python27
C:\Program Files\nodejs
C:\Program Files (x86)\iojs
C:\Program Files\iojs
C:\Users\appveyor\AppData\Roaming\npm
C:\Program Files\Microsoft SQL Server\130\Tools\Binn\
C:\Program Files (x86)\MSBuild\14.0\Bin
C:\Tools\NuGet
C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow
C:\Program Files\Microsoft DNX\Dnvm
C:\Program Files\Microsoft SQL Server\Client SDK\ODBC\130\Tools\Binn\
C:\Program Files (x86)\Microsoft SQL Server\130\Tools\Binn\
C:\Program Files (x86)\Microsoft SQL Server\130\DTS\Binn\
C:\Program Files\Microsoft SQL Server\130\DTS\Binn\
C:\Program Files (x86)\Microsoft SQL Server\110\DTS\Binn\
C:\Program Files (x86)\Microsoft SQL Server\120\DTS\Binn\
C:\Program Files (x86)\Apache\Maven\bin
C:\Python27\Scripts
C:\Tools\NUnit3
C:\Program Files\Mercurial\
C:\Program Files\LLVM\bin
C:\Program Files\dotnet\
C:\Program Files\erl8.3\bin
C:\Tools\curl\bin
C:\Program Files\Amazon\AWSCLI\
C:\Program Files\Git\cmd
C:\Program Files\Git\usr\bin
C:\Program Files (x86)\Microsoft SQL Server\140\DTS\Binn\
C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\SQLDB\DAC\140
C:\ProgramData\chocolatey\bin
C:\Program Files (x86)\Yarn\bin
C:\Program Files\Microsoft Service Fabric\bin\Fabric\Fabric.Code
C:\Program Files\Microsoft SDKs\Service Fabric\Tools\ServiceFabricLocalClusterManager
C:\Program Files (x86)\nodejs\
C:\Users\appveyor\AppData\Local\Yarn\bin
C:\Users\appveyor\AppData\Roaming\npm
C:\Program Files\AppVeyor\BuildAgent\
```

deivid-rodriguez:
Regarding path, yeah, seems reasonable. :+1: 

Regarding `fast_finish`, I think it brings more good than harm, whatever activity the repo has. Let's add it back, yeah.

MSP-Greg:
Let's say the current trunk issue has a PR/commit and it fixes it, but the same PR/commit causes an issue on Ruby22.  With `fast_finish`, you'll never see the fact that the issue was fixed in trunk.

Whatever you prefer...

deivid-rodriguez:
> Let's say the current trunk issue has a PR/commit and it fixes it, but the same PR/commit causes an issue on Ruby22. With fast_finish, you'll never see the fact that the issue was fixed in trunk.

That's so cumbersome. Never happened, and I bet it never will. I don't recall a backport breaking things in `byebug`, normally they fix things :smile:. Even if that happened, backported commits usually land in master much earlier, so it's very likely that I would've found out that it fixes things in master before it's backported.

Not sure how `fast_finish` works in Appveyor but if it works like in Travis, all entries get run in any case. The only difference is that the "failed" status is sent earlier. In any case, even if it completely stops building, I still prefer it. You can always add it back temporarily if you want to see the status of every entry.

Let's add it back.

deivid-rodriguez:
> Let's say the current trunk issue has a PR/commit and it fixes it, but the same PR/commit causes an issue on Ruby22. With fast_finish, you'll never see the fact that the issue was fixed in trunk.

If a PR is trying to fix an issue with ruby trunk, PR author will probably test locally if it fixes it. If author wants to try the fix directly in appveyor, and does not care that the "fix" breaks things in old versions, she can always temporarily remove `fast_finish: true`. I'm fine with not easily supporting that kind of workflow.

MSP-Greg:
Done.  part of the reason I removed it is that I saw travis wasn't set that way.  With `fast_finish`, you'll only see one failed job...

deivid-rodriguez:
Yeah, I was planning to add it, you reminded me of doing it :)

As I explained, I don't think that's how `fast_finish` works. The only difference is that you don't need to wait for all entries to finish before the PR is marked as "failed". At least that's how I recall it working, and that's what I understand from [this](https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/).

deivid-rodriguez:
Confirmed, #347 already has a green status, but allowed failures are still running :)

MSP-Greg:
Maybe they are a little different.  From [appveyor.yml reference](https://www.appveyor.com/docs/appveyor-yml)

```
fast_finish: true     # set this flag to immediately finish build once one of the jobs fails.
```

I think it seems that with travis, Fast-Finish determines when the job is considered done and green, in that allowed failures, even if not complete, don't hold up the green signal.

Appveyor seems to imply that the whole build finishes/stops on the first `non allow_fauilure` failure...

deivid-rodriguez:
Yes, that's what I meant by

> Not sure how fast_finish works in Appveyor but if it works like in Travis, all entries get run in any case. The only difference is that the "failed" status is sent earlier. Whatever the case, even if it completely stopped building, I still prefer it.

I'm surprised though that they copied the name but changed the behavior... :thinking: 

Thanks for your contribution @MSP-Greg!

PR #373, opened 2017-11-19
Fixup the testsuite fails on macOS

bak1an:
Default tmpdir on macOS is is a symlink pointing somewhere in
`/private/...`.  Because of this in the testsuite we should
always use realpaths since byebug outputs those by default.

Without the fix it is full of failures like https://gist.github.com/bak1an/4f6bdbbadfe96c53bd9d16dd86a87779

The PR also disables `WhereWithNotDeeplyNestedPathsTest` for `darwin` since default `$TMPDIR` on macOS nowadays is something like `/var/folders/5g/s0v3cc2x34ddw3ydhyvkryvr0000gn/T/` (which has realpath `/private/var/folders/5g/s0v3cc2x34ddw3ydhyvkryvr0000gn/T/`). This is the opposite of `not deeply nested` and just fails.

### DIFF
--- a/test/commands/where_test.rb
+++ b/test/commands/where_test.rb
@@ -115,9 +115,10 @@ module Byebug
   # nested.
   #
   # @note We skip this tests in Windows since the paths in this CI environment
-  #   are usually very deeply nested.
+  #   are usually very deeply nested and on OS X where tmp path is always
+  #   deeply nested.
   #
-  unless /cygwin|mswin|mingw/ =~ RUBY_PLATFORM
+  unless /cygwin|mswin|mingw|darwin/ =~ RUBY_PLATFORM
     class WhereWithNotDeeplyNestedPathsTest < WhereStandardTest
       def test_where_displays_current_backtrace_w_shorpaths_if_fullpath_disabled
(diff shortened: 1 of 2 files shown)


bak1an:
It fails there with a new issue now. Will take a look into that.

bak1an:
Appveyor build fix went into #376

bak1an:
Rebased that to include windows build fix

deivid-rodriguez:
Thanks @bak1an!

I have another little request. I used to run Travis against both macOS and Linux in the past but I disabled it because macOS was hard to get passing and I didn't feel it was useful. However, most developers run macOS these days so it seems like a great thing to ensure that the tests pass in that environment.

Can you try enabling Travis on macOS and see if it passes? Ref: https://docs.travis-ci.com/user/multi-os/.

bak1an:
Makes sense @deivid-rodriguez 

Will try and see what happens.

deivid-rodriguez:
I'm going to merge this for now, we can take care of adding macOS to the CI in a separate PR!

Thanks @bak1an!!

bak1an:
Thanks @deivid-rodriguez!

Most likely will spend time on macOS travis configuration next week.

PR #380, opened 2017-11-27
Change trunk download from BinTray to Appveyor

MSP-Greg:
I moved the build from BinTray to Appveyor


deivid-rodriguez:
Thanks!

MSP-Greg:
@deivid-rodriguez

Sorry for not changing it earlier.  Could you please remove [appveyor.yml:21-22](https://github.com/deivid-rodriguez/byebug/blob/master/appveyor.yml#L21-L22)?

```
  - bundler --version
  - where bundler
```
With the next line installing bundler, they don't add any info.  Also, it seems that bundler may be removing 'bundler' and only allowing 'bundle', at least as far as CLI is concerned.

Also, I added a cache, which should speed up Appveyor CI.  But, the cache size is limited by account, so if you have other repos that are also caching.  I think the limit is 1 gig, don't know if that's compressed or not...

Thanks, Greg

PR #381, opened 2017-11-27
Remove lines ruby trunk doesn't like

deivid-rodriguez:
/cc @MSP-Greg 

Can you review?


deivid-rodriguez:
No problem, I'm just too used to not committing directly to master! :)

PR #382, opened 2017-12-01
Setup osx travis builds

bak1an:
Adding macOS builds into travis matrix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ coverage
 .bundle
 .byebugrc
 .byebug_history
+.envrc
 
 lib/byebug/byebug.so
 lib/byebug/byebug.bundle
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 ---
 
-sudo: required
-
-dist: trusty
+os:
+  - linux
+  - osx
 
 cache:
   directories:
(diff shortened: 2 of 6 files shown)


deivid-rodriguez:
@bak1an Have a look at #383, I think it should make this easier!

deivid-rodriguez:
Let's rebase this and see what happens.

bak1an:
Not sure if `clang-format` is included in travis osx image, let's see how it fails.

Also I'm not really sure about building non-clang + clang in macos since clang is default there:
```
$ which cc
/usr/bin/cc
$ cc --version 
Apple LLVM version 9.0.0 (clang-900.0.38)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

We can always `brew install gcc` there to test gcc rubies in osx but don't really see any benefit in doing this 🤔

deivid-rodriguez:
I added the multiple compiler thing because I was using `gcc` myself but users compiling their rubies with `clang` would get installation errors that I hadn't seen.

But I think it's enough to do that only in Linux. :+1: Can you easily skip those entries in macOS?

bak1an:
I believe this can be done with `matrix.exclude`, will try and see.

bak1an:
Now there is a clang-format difference between linux and macos builds with braces after enum - https://github.com/deivid-rodriguez/byebug/pull/382/commits/6b72a748edeb4e470b09e1bad3c2802297185d85 🤦‍♂️

 And it looks like macos version breaks line after enum according to `BraceWrapping.AfterEnum = true`: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

@deivid-rodriguez Which version do you have? At macos it is `clang-format version 6.0.0 (tags/google/stable/2017-11-14)`

deivid-rodriguez:
Uggh. That seems like a bug in my `clang-format` version. However, I'm not sure how to install such a recent version in Linux... :S. Can you try `brew install clang-format@3.8`?

deivid-rodriguez:
Looks great @bak1an. Just one thing, have a look [at the matrix](https://travis-ci.org/deivid-rodriguez/byebug/builds/312010667), it seems like the osx config is not being picked up properly since the custom image is only being used for a single ruby version. However, I'm not sure why, your configuration looks correct to me...

bak1an:
Yeah, have noticed that too. Looks like travis doc is somewhat misleading on this. Will experiment / google for proper solutions.

deivid-rodriguez:
tears of joy

PR #383, opened 2017-12-02
Move to clang-format for C code formatting

deivid-rodriguez:
Indent seems pretty much abandoned and like a less powerful tool for the job.

This should make #382 easier.


deivid-rodriguez:
Yeah, and in the case of Travis + Linux, no other config is needed, it's available by default!

PR #386, opened 2017-12-11
Detect timeouts and other test improvements

deivid-rodriguez:
Currently appveyor build against `ruby-head` deadlocks. That means that we have a bug there, but also that appveyor takes about an hour to send back to green status to GitHub PR's (due to https://github.com/appveyor/ci/issues/1960).

I don't expect byebug test run to ever take longer than 3 minutes, so let's detect those situations and fail early.

This PR also adds some little tweaks to the tests.

Credits to @MSP-Greg for the idea and implementation!

### DIFF
--- a/script/minitest_runner.rb
+++ b/script/minitest_runner.rb
@@ -19,10 +19,16 @@ class MinitestRunner
 
     flags = ["--name=/#{filtered_methods.join('|')}/", ENV['TESTOPTS']]
 
-    Minitest.run(flags + $ARGV)
+    run_with_timeout(flags)
   end
 
   private
+
+  def run_with_timeout(flags)
+    Timeout.timeout(180) { Minitest.run(flags + $ARGV) }
+  rescue Timeout::Error
+    false
+  end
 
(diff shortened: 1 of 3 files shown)


MSP-Greg:
@deivid-rodriguez 

I'm a little sensitive re Windows/Appveyor.  Travis was also deadlocking, but its timeout is 10 minutes.

I did check, at present, there is no way to change the time setting on Appveyor.  I'll suggest it.

Re windows, thank you for considering it.  Some repos/gems have a different attitude...   Greg

deivid-rodriguez:
> I'm a little sensitive re Windows/Appveyor. Travis was also deadlocking, but its timeout is 10 minutes.

Yeah, detecting this is helpful for both CI's, but the situation was much more annoying for Appveyor.

> I did check, at present, there is no way to change the time setting on Appveyor. I'll suggest it.

:+1:

> Re windows, thank you for considering it. Some repos/gems have a different attitude... Greg

No problem, thanks for your contributions!

PR #387, opened 2017-12-11
Fix travis caching

deivid-rodriguez:
Also use the folder bundler 2 will use by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ sudo: required
 
 cache:
   directories:
-    - vendor/bundle
+    - .bundle/gems
 
 language: ruby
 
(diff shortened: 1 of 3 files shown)


MSP-Greg:
@deivid-rodriguez

Not sure if you noticed, looks like the 3 minute limit is no longer?

deivid-rodriguez:
Sorry? I didn't understand the question.

MSP-Greg:
I'm not sure what branch / PR is where, but looking at [Appveyor  history](https://ci.appveyor.com/project/deivid-rodriguez/byebug/history) shows the last build taking an hour?

Similar with Travis.

Sorry - Just noticed that the time PR hasn't been committed yet...

Greg

deivid-rodriguez:
Ahh, yes. It should start working faster as soon as I merge it. (Travis will still be slow though due to macOS limited availability... :disappointed:)

PR #393, opened 2017-12-23
show valid break point lines

ko1:
This PR is for usability. Show valid break point lines when `break` command fails.

Like that:

```
$ bin/byebug ~/src/rb/t.rb --
/home/ko1/src/rb/byebug/lib/byebug/helpers/bin.rb:15: warning: Insecure world writable dir /home/ko1/src/rb/byebug in PATH, mode 040777

[1, 10] in /home/ko1/src/rb/t.rb
    1:
=>  2: p 0
    3:
    4: def foo
    5:   if false
    6:     return 1
    7:   else
    8:     return 2
    9:   end
   10: end
(byebug) b 5
*** Line 5 is not a valid breakpoint in file /home/ko1/src/rb/t.rb.
Valid break points are:
[B]  2: p 0
     3:
[B]  4: def foo
     5:   if false
     6:     return 1
     7:   else
[B]  8:     return 2
     9:   end
[B] 10: end
    11:


(byebug)
```

It is hard to find out which lines are valid or not.
I'm not sure it is correct way to output messages, but please consider this idea.

Thanks,
Koichi


deivid-rodriguez:
@ko1 Thanks so much, this is a great usability improvement indeed! I added some extra changes so that byebug linters and tests are happy, but it looks great. This will be released with the next version! :heart:

deivid-rodriguez:
I'll probably add a few extra changes so that the surrounding code respects the user's `listsize` setting, but this is going in anyways!

stevewi:
This probably sounds like a dumb question...  I've cloned the latest byebug-master from GitHub.  How do I build/install it?  I'm used to 'gem install' but that's about it...

TIA.

MSP-Greg:
@stevewi

There may be an easier way, but normally I do the following, which does give you a gem file in case whatever...

From the repo folder:

1. `gem build byebug,gemspec`
2. `gem install -l byebug-9.1.0.gem`  (version may change)

Reference - [RubyGems 2.7.4 Command Reference](https://msp-greg.github.io/ruby_trunk/rubygems/file.command-reference.html#gem-install), same info several places on the web...

deivid-rodriguez:
@stevewi Does your application have a `Gemfile`? In that case, you can also add byebug to that file specifying this: `gem "byebug", git: "https://github.com/deivid-rodriguez/byebug"`, and then run `bundle install`. Then run any command inside your application folder prefixed with `bundle exec`.

stevewi:
My app does have a Gemfile but I also use byebug to debug one-off's and test jigs.  I installed the current master as a local gem using the method MSP-Greg described above.  When all this gets resolved, I'll install byebug as a global gem again.

The only other thing I had to change on one system is a modification to my PATH on one Linux system to include $HOME/.gems/bin.  My problem, not yours :-)

Thanks!

Issue #345, opened 2017-08-23
Add ruby 2.5 support


deivid-rodriguez:
## Problem description

Byebug doesn't yet fully support ruby 2.5.

## Expected behavior

Full tests pass and byebug works.

## Actual behavior

As of today, there's two failures in the test suite [against ruby-head](https://travis-ci.org/deivid-rodriguez/byebug/jobs/267106322), and @MSP-Greg [reports](https://github.com/deivid-rodriguez/byebug/pull/343#issuecomment-324384503) that standalone `byebug` fails to start against it. 

## Steps to reproduce the problem

Install a build of trunk's ruby and run byebug against it. Using https://github.com/deivid-rodriguez/ruby-bisect against

https://github.com/deivid-rodriguez/byebug/blob/a37248a48842cf4d082c40f38b306cd7d9860bd6/script/ci.sh#L1-L6

could be useful to find out when and how the problem was introduced.

MSP-Greg:
Thanks for links,etc.  Re bisect, build & install for ruby takes a bit of time.  I'll search thru your travis logs and see if I can narrow it down to a smaller range.  A quick look showed that it passed with a trunk build as of 25-May, and I'm sure I can narrow further.

If you have a look at my [MinGW test results page](https://msp-greg.github.io/file.mingw_test-all.html) (which isn't a log of **all** my tests), for every date listed, I've got the full build / install  saved.  I may also have additional builds not listed.

Bottom line, I can narrow it down to the point where we/you will know where the issue is, or I can run a bisect on a smaller set...

deivid-rodriguez:
No, `byebug` has no dependencies.

MSP-Greg:
Sorry, noticed that afterward.

FYI, ruby has had an mswin (Visual C) build in their CI testing for a while, but they only recently added full testing.  MinGW builds, which is what the vast majority of windows users have for Ruby, are not done by them.  Currently, I've got a few patches that are needed, etc.  IOW, someday, I may create a bisect script, but it will take a bit of dicing & chopping...

But, as mentioned, I've got a lot of saved builds.  At present, I'm between:
```
PASSES  ruby 2.5.0dev (2017-06-14 trunk 59096) [x64-mingw32]
FAILS   ruby 2.5.0dev (2017-06-24 trunk 59168) [x64-mingw32]
```

I've got a script I'm running, testing on runner_test.rb.  I haven't yet tested from command line.

[...]

PR #397, opened 2017-12-24
Fix how to detect potential_lines

yui-knk:
On Ruby 2.5, format of `RubyVM::InstructionSequence#disasm`
will be changed. By this change we can not detect which
line is valid to set breakpoints. On the other hand
Ruby 2.5 will introduce new methods

* `RubyVM::InstructionSequence#each_child`
  Iterate all direct child instruction sequences.
* `RubyVM::InstructionSequence#trace_points`
  Return trace points in the instruction sequence.

see: https://github.com/ruby/ruby/blob/9fe26ef8de1ae047a4d775ce74878a0aeb01a19b/NEWS

We can detect potential_lines by these methods.
So this commit defines new methods and call each_child &
trace_points if these methods are implemented.

Thanks to Koichi Sasada (ko1) for investigation, implemention of
new methods for Ruby and concept implementation for byebug.

This PR is a part of Ruby 2.5 support. I will send another PR to fix some
test cases.

Related: https://github.com/deivid-rodriguez/byebug/issues/345

### DIFF
--- a/lib/byebug/breakpoint.rb
+++ b/lib/byebug/breakpoint.rb
@@ -49,9 +49,27 @@ module Byebug
     #
     def self.potential_lines(filename)
       name = "#{Time.new.to_i}_#{rand(2**31)}"
-      lines = {}
       iseq = RubyVM::InstructionSequence.compile(File.read(filename), name)
 
+      if iseq.respond_to?(:each_child)
+        potential_lines_with_trace_points(iseq, {})
+      else
+        potential_lines_without_trace_points(iseq, {})
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


yui-knk:
On my local environment

```shell
$ ruby -v
ruby 2.5.0dev (2017-12-24 trunk 61432) [x86_64-darwin14]

$ bundle exec rake test
465 runs, 528 assertions, 7 failures, 0 errors, 3 skips
```

deivid-rodriguez:
Looking great, thanks so much! I'll merge as soon as CI is done!

PR #398, opened 2017-12-24
Fix tests on Ruby 2.5

yui-knk:
On Ruby 2.5, some location of
`rb_debug_inspector_backtrace_locations` may be changed.
This change affects the line number of current frame
when breakpoint is set to method. I think this change will not
affect byebug users because after this change the line number of
current frame is still inside of target method.
So, this commit adds branches to test cases to fix tests on Ruby 2.5.

Related: https://github.com/deivid-rodriguez/byebug/pull/397 and https://github.com/deivid-rodriguez/byebug/pull/345

This PR will fix all test cases on all target ruby versions.


deivid-rodriguez:
Thanks @yui-knk!

I agree it shouldn't be a big deal for `byebug` users, but I still would like to know more about the change. Was it a compromise for some other improvements? Was it because the new locations are considered better? What is the new behavior with empty methods (for example, an empty Rails controller action)?

yui-knk:
> I still would like to know more about the change

I will write more detailed information about the change.

> What is the new behavior with empty methods

Nice catch!! Thanks!!
Just now ko1 fixed the new behavior not to break compatibility with Ruby 2.4.

https://svn.ruby-lang.org/cgi-bin/viewvc.cgi?view=revision&revision=61457

By the way, I notice byebug has no test cases for empty methods.
So I create PR to add test cases.
https://github.com/deivid-rodriguez/byebug/pull/399

yui-knk:
@deivid-rodriguez 

> I still would like to know more about the change

@ko1 will write more detailed information about the change :)

ko1:
> Was it a compromise for some other improvements?

For performance. Ruby 2.4 uses the `trace` instruction and the beginning of method line number is represented by `trace` instruction. On Ruby 2.5, `trace` instruction is removed and the beginning of line info is actual code line (effective code line).

> Was it because the new locations are considered better?

No. Implementation's choice. "compatibility" vs. "performance" and I believe this change is reasonable for this trade-off.

deivid-rodriguez:
Thanks @ko1, sounds good!

deivid-rodriguez:
@yui-knk Hope you don't mind a force-pushed to this branch to rebase it and start testing against ruby 2.5.0. :pray:

deivid-rodriguez:
@ko1 Final questions about this change:

In https://bugs.ruby-lang.org/issues/14104#Compatibility-issues you mention two compatibility issues:

> * Line numbers on :return/:b_return events show the last executed lines, instead of end lines (without return statement).
> * begin block doesn't show :line events.

Regarding the first one, it was reverted in https://github.com/ruby/ruby/commit/919fa894f5e0fed105ad1513cb8be0748ae46295, right? Did this mean sacrificing any performance improvement for the general, non-debugging, case? I hope not... :sweat_smile:.

The second one I assume it stays. However you added compatibility for "empty constructions" in https://github.com/ruby/ruby/commit/360c00572feb09da9d1575f667348773acfdcefd, so I wonder if it'd be possible to do something similar in the general case (again, without sacrificing performance gains, of course).

Whatever the case, this is ready now and I'll be releasing 2.5 support with `byebug 10.0.0` very soon! :)

ko1:
> Regarding the first one, it was reverted in ruby/ruby@919fa89, right? Did this mean sacrificing any performance improvement for the general, non-debugging, case? I hope not... 😅.

No. No performance penalty.

> The second one I assume it stays.

Yes.

> However you added compatibility for "empty constructions" in ruby/ruby@360c005, so I wonder if it'd be possible to do something similar in the general case (again, without sacrificing performance gains, of course).

We need performance penalty (1 nop operation for such case). Do you use it?

deivid-rodriguez:
Aha, no, I was just wondering because I guess it would be less surprising for the user that breakpoints on methods always stop in a consistent location.

But the first "effective" line (or the method definition line if the method is empty) is definitely a good enough and reasonable trade-off!

Thanks so much for your work, @ko1 & @yui-knk!

PR #399, opened 2017-12-25
Add breakpoints test cases for empty methods

yui-knk:
When stopped by breakpoint for an empty method
whose body is empty, the line of `def method_name`
would be displayed.

### DIFF
--- a/test/commands/break_test.rb
+++ b/test/commands/break_test.rb
@@ -109,6 +109,56 @@ module Byebug
   end
 
   #
+  # Tests adding breakpoints to empty methods
+  #
+  class BreakAtEmptyMethodsTest < TestCase
+    def program
+      strip_line_numbers <<-RUBY
+          1:  module Byebug
+          2:    #
+          3:    # Toy class to test breakpoints
+          4:    #
+          5:    class #{example_class}
+          6:      def self.a
+          7:      end
+          8:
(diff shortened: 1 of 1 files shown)


test/commands/break_test.rb
```diff
@@ -108,6 +108,46 @@ def test_setting_breakpoint_to_invalid_location_shows_an_error
...
+          1:  module Byebug
+          2:    #
+          3:    # Toy class to test breakpoints
+          4:    #
+          5:    class #{example_class}
+          6:      def self.a
+          7:      end
+          8:
+          9:      def b
+         10:
```
deivid-rodriguez: Is the extra empty line intentional?

deivid-rodriguez:
Thanks @yui-knk! Added a couple of comments, it's great to add this edge cases to the test suite!

test/commands/break_test.rb
yui-knk: Yes it is intentional to test lineno 10 is not displayed.

deivid-rodriguez: :+1:

yui-knk:
@deivid-rodriguez I added one-line method case and some comments about why `#b` has extra empty line.

yui-knk:
❤️

PR #420, opened 2018-01-07
Use latest rubies on TravisCI configuration

deivid-rodriguez:


deivid-rodriguez:
Closed in favor of #421.

PR #425, opened 2018-01-12
Tweak appveyor

deivid-rodriguez:
Some tweaks to our appveyor config.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,15 +1,16 @@
 ---
-init:
-  # clean path up to just what we need
-  - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\Git\cmd;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Windows\system32;C:\Windows
-  # Loads trunk build, updates MSYS2 / MinGW to most recent gcc compiler, replaces GDBM & OpenSSL packages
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@deivid-rodriguez

Thanks for removing 'fast finish'.  Along with being a ruby 'windows' PITA, I also have a doc site with Ruby, Rails, AWS, and a few popular gems.  It's [here](https://msp-greg.github.io/).  Anyway, there have been times when trunk passes all the ruby tests, but caused issues with the doc system (based on YARD with a lot of change).  Hence, with 'fast finish', one might not see those issues.  That's also why I often start from the newest ruby (trunk) and work back.  You can see if trunk fails quickly, regardless of whether it's included in 'allow_failures'.

From what I've seen, with ruby versions from 2.0 forward (excepting trunk), if one fails. they all fail, as most repos have gotten past the 2.4 changes.  Even better if restricted to 2.2 forward.

Re trunk, at present, Lars' trunk build (RubyInstaller2) is done once a day, mine is done three times a day, and mine is fully tested (which has been what I've concentrated on). Lars could change both things  relatively easily.

Anyway, the way mine is set up, a build is only available for use if it builds & passes all tests.  I do not do scheduled 32 bit builds (x86), but I do them occasionally locally.  I might be mistaken, but I doubt there are many windows users running 32 bit, at least with trunk.

PR #426, opened 2018-01-13
Drop TravisCI and building on macOS

deivid-rodriguez:
It's too slow to be useful rather than annoying...


MSP-Greg:
Does CircleCI have head available for testing?  Asked by a windows guy, but I didn't see it.  Maybe when Travis is back to normal add head as before (not using MacOS)?

[Travis status](https://www.traviscistatus.com/#month) shows their 'MacOS builds' max dropped from 180 to 90 recently.  Haven't looked to see if they explain why...

bak1an:
It looks like travis just does not have enough capacity to process all the macos builds. Note that linux builds backlog is empty on their status page and macos backlog is never empty.

CircleCI seems to have macos builds in paid only plans. They offer some [free builds for open source](https://circleci.com/pricing/#faq-section-os-x) though, could be an option.

MSP-Greg:
>  Note that linux builds backlog is empty on their status page

That's why I linked to the monthly.  I needed linux builds the last week of the year.

I'm just pointing out that maybe Travis should do a linux head build and comment out the MacOS until the backlog disappears...

bak1an:
@MSP-Greg  Looks like travis is doing macos builds only at the moment, linux builds are on circleci.

@deivid-rodriguez Another way to handle this might be to just ignore macos builds and consider pull requests green as soon as windows and linux builds are green. Macos ones will finish someday, so at least if they are not passing we will see it as a red readme badge.

bak1an:
Month status page indeed shows that there were times when macos build were not queuing at travis.

MSP-Greg:
@bak1an 

> Looks like travis is doing macos builds only at the moment, linux builds are on circleci.

Yes, but I don't see a ruby trunk/head build done on CircleCI.  My point is that maybe one should be done on Travis?

deivid-rodriguez:
@bak1an I think your idea is a good compromise, I'll set the TravisCI check as non required and rely on the badge to found out about breakage :+1:

By the way I'll on holidays for one or two weeks so I won't be very responsive.

deivid-rodriguez:
@MSP-Greg I don't mind adding `ruby-head` to TravisCI, specially now that we'll be no longer requiring it for PR's status. Adding it to CircleCI is a different beast since it would require either building a base docker image against ruby-head and push it to dockerhub X times a day (maybe someone is already doing this?),  or doing it on the fly for every PR. It sounds like a bit of work to me but feel free to give it a try if you're up to it!

deivid-rodriguez:
Back from holidays! I've done as suggested by @bak1an and marked Travis CI checks as non required!

Issue #360, opened 2017-10-13
NoMethodError (undefined method `step_out' for #<Byebug::Context:0x00000003c62f48>)


issei126:
## Problem description

I get the error. The error message is on title. 
I found the same error in https://github.com/deivid-rodriguez/byebug/issues/200, but the error was not resolved.
The difference is I use Docker on mac os.

Rails 5.0.2
Ruby 2.4.1
byebug 9.1.0
Docker Version 17.09.0-ce-mac35 (19611)
mac os 10.12.6

issei126:
I Reproduce the error.

Dockerfile
```
FROM ruby:2.4.1
RUN apt-get update -qq && \
 apt-get install -y build-essential libpq-dev curl && \
 curl -sL https://deb.nodesource.com/setup_4.x | bash - && \
 apt-get install -y nodejs && \
 mkdir /app
WORKDIR /app
ADD Gemfile /app/Gemfile
ADD Gemfile.lock /app/Gemfile.lock
RUN bundle install
ADD . /app
```

Gemfile
```
source 'https://rubygems.org'


# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '5.0.1'
gem 'sqlite3'
# Use sqlite3 as the database for Active Record
group :development, :test do
  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
  gem 'byebug'
  gem 'debase', "0.2.1"
  gem 'ruby-debug-ide', "0.6.0"
end
```

application code
```
class TopController < ApplicationController
  def index
    byebug
  end
end
```

Then, I try to access `localhost:3000/top/index`, The error blow raises.
```
Completed 500 Internal Server Error in 464ms (ActiveRecord: 0.0ms)



NoMethodError (undefined method `step_out' for #<Byebug::Context:0x007fa4608278d0>):

app/controllers/top_controller.rb:3:in `index'
```

Please give me an advise.
Thanks.

tomaszwojcik:
Same issue here, my specs:

* Rails 5.0.4
* Ruby 2.4.1p111
* byebug 9.1.0
* Docker Version 17.09.0-ce-mac35 (19611)
* macOS 10.13.1 (17B48)

deivid-rodriguez:
Sorry I never got to answer this. `Byebug` is incompatible with `debase`.

ducnguyenhuy:
I got the same issue and I'm not using debase gem. Any suggestion please?

Issue #200, opened 2015-12-29
NoMethodError (undefined method `step_out' for #<Byebug::Context:0x00000003c62f48>)


benams:
I get this error everywhere I use 'byebug' in my controller. I couldn't find any information about this message, but it's very weird as I've used byebug successfully until now. 
After the error message I can see a very long and not informative trace which is related to active_record(?!), and then the server crashes and killed.
* I use rails 4.2.1 with byebug 8.2.1.

deivid-rodriguez:
Hi @benams, sorry I forgot about this ticket. Are you still experiencing the problem? It looks like an installation problem...

benams:
sorry but it was a long time ago, and I got a few more issues with the gem, so I replaced it with "pry"

deivid-rodriguez:
Cool, closing then! Thanks @benams.

PR #443, opened 2018-03-21
Rename function to prevent conflict to debase

tzmfreedom:
When we use byebug and debase gem simultaneously, there are conflict because extern function name is conflict each other.

For example, with byebug and debase gem, if we use `byebug` method to debug code, the following error is occured (this is the same issue as #200 #360 )
```
NoMethodError (undefined method `step_out' for #<Byebug::Context:0x007fa4608278d0>)
```

This is because both gem have extern function `Init_context` and `context_create` and if we call these function in byebug gem, it calls debase's function. (Maybe other development gem use these function name, because `context` is commonly used variable name)

This PR solve by renaming function.

### DIFF
--- a/ext/byebug/byebug.h
+++ b/ext/byebug/byebug.h
@@ -125,7 +125,7 @@ extern VALUE next_thread;
 
 /* functions from context.c */
 extern void Init_context(VALUE mByebug);
-extern VALUE context_create(VALUE thread);
+extern VALUE byebug_context_create(VALUE thread);
 extern VALUE context_dup(debug_context_t *context);
 extern void reset_stepping_stop_points(debug_context_t *context);
 extern VALUE call_with_debug_inspector(struct call_with_inspection_data *data);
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Makes sense to me @tzmfreedom! I guess you could potentially bundle both debuggers (for example, `debase` for `Rubymine`, and `byebug` for direct debugging) and things should work. Can you add a change log entry for this under `### Fixed`? I'll release `byebug 10.0.1` with this fix.

ext/byebug/byebug.h
```diff
@@ -125,7 +125,7 @@ extern VALUE next_thread;
 
 /* functions from context.c */
 extern void Init_context(VALUE mByebug);
```
deivid-rodriguez: You need to replace this function as well, right?

tzmfreedom: Thank you for your review. I fixed it.

deivid-rodriguez:
Thank you @tzmfreedom!

PR #444, opened 2018-03-21
Release 10.0.1

deivid-rodriguez:
Getting ready for 10.0.1 release.

/cc @tzmfreedom.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 10.0.1 - 2018-03-21
+
 ### Fixed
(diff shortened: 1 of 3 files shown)


greysteil:
@deivid-rodriguez - so sorry for the cross-reference spam above! FYI, I only realised it was happening today (was due to a change we made last week) and fixed it immediately in https://github.com/dependabot/dependabot-core/commit/2bd343b0ef8638ff15914af155c96db0b6bd231a.

deivid-rodriguez:
Hadn't even noticed! No problem, of course, and thanks again for your work!

PR #447, opened 2018-03-26
Rename functions that conflict with debase's

tzmfreedom:
On my last pullrequest (#443), I fixed conflict with debase by renaming context initializer functions(e.g. Init_context() => Init_byebug_context()).

But I missed to fix other functions to conflict with debase's, for example Init_breakpoint(). So breakpoint isn't worked when using with debase.

I fixed it on this PR. On this PR, we can use byebug with debase completely.

Fixed point(rename) is the following

* Init_breakpoint() => Init_byebug_breakpoint()
  * conflict to https://github.com/denofevil/debase/blob/c317d87bcf48a66f25da858a9ab6d7f4c5486ee7/ext/debase_internals.h#L110
* reset_stepping_stop_points() => byebug_reset_stepping_stop_points()
  * conflict to https://github.com/denofevil/debase/blob/c317d87bcf48a66f25da858a9ab6d7f4c5486ee7/ext/debase_internals.h#L79
* add_to_locked() => byebug_add_to_locked()
* pop_from_locked() => byebug_pop_from_locked()
* remove_from_locked() => byebug_remove_from_locked()
  * conflict to https://github.com/denofevil/debase/blob/c317d87bcf48a66f25da858a9ab6d7f4c5486ee7/ext/debase_internals.h#L93-L95

I think extern function name is unique globally as much as possible because C doesn't have namespace to function name.
For example, on msgpack-ruby(https://github.com/msgpack/msgpack-ruby), there is some functions that is prefixed with `MessagePack` or `msgpack`.

So ideally we may rename all extern function name by adding prefix(e.g. byebug_xxx()).
But I fixed only functions that conflict with debase now.
(I don't know the reason, some functions is the same as debase's.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Error when using byebug with `debase` gem (#447, @tzmfreedom)
+
 ## 10.0.1 - 2018-03-21
 
 ### Fixed
--- a/ext/byebug/breakpoint.c
+++ b/ext/byebug/breakpoint.c
@@ -491,7 +491,7 @@ find_breakpoint_by_method(VALUE breakpoints, VALUE klass, ID mid, VALUE bind,
 }
 
 void
-Init_breakpoint(VALUE mByebug)
+Init_byebug_breakpoint(VALUE mByebug)
 {
   breakpoint_max = 0;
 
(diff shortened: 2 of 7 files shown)


deivid-rodriguez:
Thank you again @tzmfreedom! Looks good to me. I'll merge and release 10.0.2 with it as soon as I fixed CI configuration :)

deivid-rodriguez:
Thanks so much for the patch. The reason for the name conflicts is that the C extension of `byebug` was initially forked from `debase` actually :)

I'll release 10.0.2 tomorrow with this fix :+1:

PR #467, opened 2018-06-12
small discrepancies in guide

RudyOnRails:
several of these lines don't match the original program, so I (think I) matched them up a bit more. not super important, but thought I'd help since byebug is my favorite tool.


RudyOnRails:
sorry about running CI on this md change - forgot.

deivid-rodriguez:
Thanks so much!

PR #469, opened 2018-06-16
Add SemVer compatibility badge

greysteil:
(Similar PR to https://github.com/deivid-rodriguez/pry-byebug/pull/181.)

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=byebug&package-manager=bundler&version-scheme=semver)

Would love your opinion - is it useful? Anything you'd change about it (e.g., keep / lose the logo)? If you click it there's an explanation of how it's calculated.

Any feedback super appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,20 @@
 [![Maintainability][mai]][mai_url]
 [![Coverage][cov]][cov_url]
 [![Gitter][irc]][irc_url]
+[![SemVer][semver]][semver_url]
 
 [gem]: https://img.shields.io/gem/v/byebug.svg
 [mai]: https://api.codeclimate.com/v1/badges/f1a1bec582752c22da80/maintainability
 [cov]: https://api.codeclimate.com/v1/badges/f1a1bec582752c22da80/test_coverage
 [irc]: https://img.shields.io/badge/IRC%20(gitter)-devs%20%26%20users-brightgreen.svg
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hi @greysteil! I love your badge!

However, I'm hesitant on how useful it is in the case of these gems. Byebug is never meant to be used on final committed code, thus no byebug's code should be run at all on apps test suites. So it seems misleading. I think the 98% here is just test suite flakiness and has nothing to do with byebug's releases being incompatible 2% of the times. Thoughts?

greysteil:
> Byebug is never meant to be used on final committed code, thus no byebug's code should be run at all on apps test suites.

That's a very good point - I guess in that case the score is a bit meaningless here! I'll close this.

The 2% is almost certainly due to flakiness, which is frustrating. I quite like how straightforward to calculation of the badge is (% of test runs that stayed green), so don't really want to mask it by making it a score out of 10 or something like that, but maybe that would be clearer for people seeing it for the first time?

deivid-rodriguez:
I think the misleading-ness due to flakyness is kind of unavoidable. But as you said, a % score is straighforward both to calculate and understand. I would keep it as it is.

greysteil:
👍 - useful feedback. Easy to second-guess these things, so good to have an outside opinion.

What about the logo? Keep or drop?

deivid-rodriguez:
I like the logo in the badge too!

PR #483, opened 2018-08-29
Remove unused codeclimate config

deivid-rodriguez:


deivid-rodriguez:
Nah, don't see a why to disable everything and just leave code coverage.

PR #498, opened 2018-10-20
Colorize output

vbrazo:
I was reading this PR https://github.com/deivid-rodriguez/byebug/pull/284 and decided to come up with a suggestion. Since we're using the `pry` gem, we could use one of its helpers to colorize the output: `Pry::Helpers::BaseHelpers.colorize_code` - [Reference](https://www.rubydoc.info/github/pry/pry/Pry/Helpers/BaseHelpers). The `pry-rails` uses this helper as well. Check [this link](https://github.com/rweng/pry-rails/blob/master/lib/pry-rails/commands/recognize_path.rb#L31) out.

I added the `ENV["GEM_ENV"]` because when adding colors to the output, a lot of tests break, and I'd like to ask you if it's really worth to change all the tests since either formats reproduce the same behaviours. I think testing in the raw version is better.

Let me know what you think about this idea.

The output should look like:

screen shot 2018-10-20 at 6 39 57 pm


deivid-rodriguez:
@vbrazo Thanks! This is an interesting idea, and the pry coloring is really pretty... But somehow it feels wrong to introduce a dependency on `pry` just for syntax highlighting... Maybe we could pull from pry on the necessary code? I'm not sure.

I actually started working on this myself a while ago using the [rouge gem](https://github.com/jneen/rouge), but I never finished it and it didn't look as pretty as pry... :disappointed:. Anyways, I rebased and pushed the WIP branch in case you're interested :point_right: https://github.com/deivid-rodriguez/byebug/compare/syntax_highlight.

deivid-rodriguez:
Actually, it looks pretty decent I think... I'll try to wrap it up!

syntaxbyebug

vbrazo:
Looks good, better than plain text. Let me know how I can help you here. I’d like to learn more about this gem.

deivid-rodriguez:
To that branch I would like to add:

* `--color/--no-color` options to the CLI.
* Set the default to `--color` if we're a TTY, to `--no-color` otherwise.
* Add some high level specs for the highlighting.

On further iterations we could add settings to show, list & set a theme from the available themes in `rouge`.

If you want to pick it up, feel free to give it a try!

vbrazo:
Awesome. I'll work on it in the next days 👍

PR #499, opened 2018-10-20
Add basic SimpleCov configurations

vbrazo:
- Remove `/test` from local reports. When opening the `coverage/index.html` in my browser, I could see the test folder there
- Add HTML formatter


deivid-rodriguez:
@vbrazo I actually include test in coverage reports intentionally. It helps me find dead tests or test helpers. Regarding the formatter, isn't HTML the default?

vbrazo:
I think you're right. Let's close this one for now.

PR #503, opened 2018-10-27
readme writing changes for readability

tamkylet:
I have made writing changes in the Readme file to improve readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Command     | Aliases      | Subcommands
 
 ## Semantic Versioning
 
-Byebug attempts to follow [semantic versioning](https://semver.org) as well as
+Byebug attempts to follow [semantic versioning](https://semver.org) and
 bump major version only when backwards incompatible changes are released.
 Backwards compatibility is targeted to [pry-byebug] and any other plugins
 relying on `byebug`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,7 +137,7 @@ Command     | Aliases      | Subcommands
 
 ## Semantic Versioning
 
-Byebug tries to follow [semantic versioning](https://semver.org) and tries to
+Byebug attempts to follow [semantic versioning](https://semver.org) as well as
```
deivid-rodriguez: Bumping major version when backwards incompatible changes are released is part of semantic versioning, so I'm not sure as well is the best fit here? This section needs to be rewritten anyways, since this is not the exact policy I've been following, but I wanted to mention this in case you can think of a better wording.

tamkylet: Thanks for the comment! I am a student as well as a Rails enthusiast and I love using Byebug.

How about:
"Byebug attempts to follow semantic versioning and bump major version only when backwards ..."?

deivid-rodriguez: Yeah, I think that reads better! :+1:

deivid-rodriguez:
Thanks so much for your contribution @tamkylet!

PR #512, opened 2018-11-23
Remove force_ruby_platform setting from appveyor

deivid-rodriguez:
I don't think it's necessary now, and it causes issues when adding dependencies.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,6 @@ install:
   - gem update --system 2.7.8
   - gem install bundler --version 1.17.1 --force
 
-  - bundle config force_ruby_platform true
   - bundle install --jobs 3 --retry 3 --path .bundle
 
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@deivid-rodriguez 

Just a quick note about why.  A pre-compiled gem may be released before a new Ruby release, but not updated for a while.  Hence it may not have pre-compiled files for the new Ruby version.  RG isn't the best at dealing with Ruby version constraints, so it may try to install the pre-compiled gem, but it will error on install.  Same for trunk...

This is something I have been working on changing in RG, but it goes to a larger issue of the coupling between the code to select which gems are installed (using install/update) with the code to actually install them.  Much of RG also uses 'tuples' (to be more 'efficient'), but tuples do not contain Ruby version constraints...

PR #406, opened 2017-12-31
Love for remote debugging

deivid-rodriguez:
Fixes #82.
Fixes #141.
Fixes #239.
Fixes #274.
Fixes #302.

Issue #274, opened 2016-05-31
Remote Debugger Hanging Up Crashes Byebug


LtSquigs:
**Problem:** If the socket for a remote connection is closed after connecting to the remote server, further writes/reads will crash the server.

**Version:** 9.05

**Steps to Reproduce**:
 1) Run this example program

```
require 'byebug'
require 'byebug/core'

Byebug.wait_connection = true
Byebug.start_server 'localhost', ENV.fetch("BYEBUG_SERVER_PORT", 1048).to_i

puts 'a'
byebug
puts 'b'
sleep 5
puts 'c'
byebug
puts 'd'
```

2) Connect to this using the remote debugger
3) When it encounters the first `byebug` statement, `continue`
4) While the program sleeps, terminate the byebug CLI (CTRL + C)
5) After the program wakes up, when it hits the second `byebug` statement, it will crash with an `Errno::EPIPE` exception

**Expectation**:  The server should either detect when the socket is closed or if an EPIPE exception has happened, and stop communicating over the socket (treat it as if there was no connection).

While the above example is very arbitrary, a more real world example of where this happens: debugging remote rails applications. If you have a byebug statement in a controller, connect to it remotely, debug the area, then the debugger closes for some reason you will have a crash.

Fixing this will also allow users to "detach" the debugger by interrupting it between byebug calls (useful if you have a constantly running dev server).

deivid-rodriguez:
@LtSquish Thanks for the great report. Want to try fixing it? 😄

deivid-rodriguez:
@LtSquigs This should be fixed by #406, feel free to give the branch a try and give feedback!

PR #517, opened 2018-12-08
Make `program_with_two_breakpoints` has exactly 2 breakpoints

yui-knk:
This test case was introduced as a regression test of
https://github.com/deivid-rodriguez/byebug/issues/274
in https://github.com/deivid-rodriguez/byebug/pull/406.

The description of issues/274 says an exception will be raised
if the byebug CLI is terminated when main program is sleeping.

Before this commit there are 2 breakpoints (`byebug` method call)
before `sleep 3` and `"cont"` is called once because the arugment of
`remote_debug_connect_and_interrupt` is `"cont"`.
I think this is not intended.

And the method name of test case is `program_with_two_breakpoints`,
on the other hand the program had 3 breakpoints (`byebug`).

### DIFF
--- a/test/remote_debugging_test.rb
+++ b/test/remote_debugging_test.rb
@@ -55,13 +55,12 @@ module Byebug
         16:
         17:    byebug
         18:    thingy = #{example_class}.new
-        19:    byebug
-        20:    thingy.a
-        21:    sleep 3
-        22:    thingy.a
-        23:    byebug
-        24:    thingy.a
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I'm unsure why CircleCI didn't trigger but I'm merging this because it makes sense. Thanks @yui-knk!

yui-knk:
Thanks for your quick response :)

PR #536, opened 2019-02-01
Add irb as a runtime dependency

bastelfreak:
prybug depends on irb: https://github.com/deivid-rodriguez/byebug/blob/d971a2b40807faf4c8654b204ddc2b30d90351fe/lib/byebug/commands/irb.rb#L4
irb isn't bundled with ruby, starting with ruby2.6. irb is a standalone
gem, so we cannot rely on it being present on the target system.

### DIFF
--- a/byebug.gemspec
+++ b/byebug.gemspec
@@ -29,5 +29,7 @@ Gem::Specification.new do |s|
   s.extensions = ["ext/byebug/extconf.rb"]
   s.require_path = "lib"
 
+  s.add_runtime_dependency "irb", "~> 1.0"
+
   s.add_development_dependency "bundler", "~> 2.0"
 end


deivid-rodriguez:
@bastelfreak `irb` is a default gem in ruby 2.6. That means it can't be removed, so I think we can assume it's present, actually.

bastelfreak:
Is is actually a default gem? At least on archlinux it isn't bundled with ruby2,6, as far as I know. It's a stand alone gem.

deivid-rodriguez:
I thought so, but maybe I'm wrong... Either that, or a packaging bug? :thinking:

deivid-rodriguez:
It seems like a packaging bug indeed: https://git.archlinux.org/svntogit/packages.git/commit/?h=packages/ruby&id=509f2d159b11686c8739154c6fe50978fbcddbb4

deivid-rodriguez:
I opened a ticket about it: https://bugs.archlinux.org/task/62709

deivid-rodriguez:
Hi @bastelfreak. archlinux has readded irb as a default dependency of their ruby package, so I think we can close this?

zenspider:
Pls close

dsandstrom:
This is unrelated to the original intent, so please let me know if I should open a new issue. However, irb is no longer a default gem in ruby 4.

https://rubylearning.com/blog/2026/03/09/ruby-4-adoption-guide/

zenspider:
@deivid-rodriguez !! cmon man... please add some maintainers to help out.

PR #553, opened 2019-04-11
Add the byebug line # to make the output clickable

senhalil:
Adding the exact line number of the byebug statement to the end of the filename on the output makes the statement clickable in IDE's and editors supporting this functionality. Currently clicking the path goes to the top of the file. With the modification, clicking takes the user to the exact byebug line in the document.

For example in vscode it looks like follows:
vscode_byebug

### DIFF
--- a/lib/byebug/commands/list.rb
+++ b/lib/byebug/commands/list.rb
@@ -113,7 +113,7 @@ module Byebug
     # @param max [Integer] Upper bound
     #
     def display_lines(min, max)
-      puts "\n[#{min}, #{max}] in #{frame.file}"
+      puts "\nStopped at #{frame.file}:#{frame.line}. Showing lines [#{min}, #{max}]:"
 
       puts source_file_formatter.lines(min, max).join
     end
--- a/test/processors/command_processor_test.rb
+++ b/test/processors/command_processor_test.rb
@@ -202,7 +202,7 @@ module Byebug
     def test_autolists_lists_source_before_stopping
       debug_code(program)
 
-      check_output_includes "[5, 14] in #{example_path}"
+      check_output_includes "Stopped at #{example_path}:11. Showing lines [5, 14]:"
     end
 
     def test_shows_error_when_current_source_location_is_unknown
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+* [#553](https://github.com/deivid-rodriguez/byebug/pull/553): Shows the line number in the byebug statement so that supporting IDEs can take cursor to the debug point instead of the head of the file ([@senhalil]).
+
 ## [13.0.0] - 2026-01-15
 
 ### Fixed
--- a/test/commands/list_test.rb
+++ b/test/commands/list_test.rb
@@ -33,7 +33,7 @@ module Byebug
         enter "list"
         debug_code(program)
 
-        check_output_includes "[12, 14] in #{example_path}"
+        check_output_includes "Stopped at #{example_path}:13. Showing lines [12, 14]:"
       end
     end
 
@@ -42,7 +42,7 @@ module Byebug
         enter "set listsize 5.0", "list"
         debug_code(program)
 
-        check_output_doesnt_include "[11, 15] in #{example_path}"
+        check_output_doesnt_include "Stopped at #{example_path}:13. Showing lines [11, 15]:"
       end
     end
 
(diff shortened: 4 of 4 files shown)


lib/byebug/commands/list.rb
```diff
@@ -113,7 +113,7 @@ def move(line, size, direction = "+")
     # @param max [Integer] Upper bound
     #
     def display_lines(min, max)
-      puts "\n[#{min}, #{max}] in #{frame.file}"
+      puts "\n[#{min}, #{max}] in #{frame.file}:#{(max+min)/2}"
```
deivid-rodriguez: I think it's simpler to write `frame.line` here?

deivid-rodriguez (changes requested):
Thanks @senhalil, I like this :+1:.

Can you rebase the PR, fix the tests and add a changelog entry under the "### Added" section explaining why we're doing this?

lib/byebug/commands/list.rb
senhalil: You are right, it is more proper like that. I will modify as follows and do the things in your other comment at the weekend.
```suggestion
      puts "\n[#{min}, #{max}] in #{frame.file}:#{frame.line}"
```

senhalil:
I realised fixing the tests is not as straight forward as I thought at the beginning. 

The complication comes from the fact that one can use `list` as many times as they want without modifying the frame.line. 

Then the output becomes as follows:
```
 "[7, 16] in /tmp/byebug_example.rb:13",
 "7:       \"%1\"",
 "8:     end",
 "9:   end",
 "10:",
 "11:   byebug",
 "12:",
 "=> 13:   str = ByebugExampleClass.new.build_percentage_string",
 "14:",
 "15:   str",
 "16: end",
 "(byebug)",
 "",
 "[6, 8] in /tmp/byebug_example.rb:13",  <--------- THIS LINE
 "6:     def build_percentage_string",
 "7:       \"%1\"",
 "8:     end",
 "(byebug)"
``` 
13 is still useful for terminal debug sessions since it tells the next line to be executed; however, then the tests needs to be modified as such:
```    
check_output_includes "[6, 8] in #{example_path}:13"
```
I don't know if you are okay with this modification.

To be honest, if someone is using `list` this means they are probably on terminal which means they don't need the "hyperlink' support; but I think giving the frame.line info still useful. What do you think?

deivid-rodriguez:
Hi @senhalil! So... It sounds that just recalculating the middle line every time like you did in the beginning is a better option after all?

senhalil:
> Hi @senhalil! So... It sounds that just recalculating the middle line every time like you did in the beginning is a better option after all?

Hi @deivid-rodriguez In my opinion, your suggestion is still the superior option, because it gives a useful info -- i.e., next instruction to be executed --, my original way didn't give any info (except average of two numbers :) ).

If someone is using `list` methods, the location of the instruction arrow might be useful to them. 

So I suggest keeping the current version if you are not oppose to it.

PS: Is it me who needs to mark "Changes requested" as resolved or you?

deivid-rodriguez:
Yeah, right.

I'm still worried that this might confuse some users into thinking this is a bug and we might get reports about it. Is there any way we can clarify the status line in this regard? Maybe something like:

> Stopped at /path/to/file:\<line\>. Showing lines [min, max]

Or something like that?

deivid-rodriguez:
> PS: Is it me who needs to mark "Changes requested" as resolved or you?

I think it's me :+1:.

senhalil:
> I'm still worried that this might confuse some users into thinking this is a bug and we might get reports about it. Is there any way we can clarify the status line in this regard? Maybe something like:

I understand.

> > Stopped at /path/to/file:<line>. Showing lines [min, max]
> 
> Or something like that?

I am happy with that wording, if you are. It is clean and clear.

zenspider:
I know it's been a while for this PR. I'd like to nudge the author into giving this another go. If not, I'd be happy to take over this branch, rebase it, and get it passing again.

I also think all the line numbers in the diff could be cleaned up into the variable which should be able to clean it up further.

senhalil:
It has been a minute! Thanks for the ping @zenspider. 

@deivid-rodriguez rebased the branch and made the change you requested.

test/processors/command_processor_test.rb
```diff
@@ -202,7 +202,7 @@ def program
     def test_autolists_lists_source_before_stopping
       debug_code(program)
 
-      check_output_includes "[5, 14] in #{example_path}"
+      check_output_includes "[5, 14] in #{example_path}:11"
```
senhalil:
```suggestion
      check_output_includes "Stopped at #{example_path}:11. Showing lines [5, 14]:"
```

deivid-rodriguez (approved):
Looks great, thank you!

PR #562, opened 2019-05-28
Fix post mortem location after exception inside `load`

### LINKED ISSUE #165: Post Mortem support broken in rubies >= 2.2.4
We're getting one test failure against latest `ruby-head` (https://travis-ci.org/deivid-rodriguez/byebug/jobs/76083918). Needs investigation.

deivid-rodriguez:
After https://github.com/ruby/ruby/commit/1998039ea4f867583d7e37ce200a88490707c330, our post mortem debugging partially broke, because ruby started emitting two `:raise` events when an exception would happen during a `load` call. In this situation, the second `:raise` event would overwrite the first one, and the context location would be incorrect, leaving the user in an unexpected location.

We fix the issue by watching for the case where we have already saved the context to the raised exception, and avoiding the overwrite.

Fixes #165.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+* [#562](https://github.com/deivid-rodriguez/byebug/pull/562): post mortem mode landing in the wrong line when coming from an exception inside a `Kernel.load` call.
+
 ### Removed
 
 * Support for MRI 2.3. Byebug no longer installs on this platform.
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Build against ruby-head seems broken, but it seems unrelated to this PR. I'll investigate it.

deivid-rodriguez:
ruby-head build was fixed in #563.

deivid-rodriguez:
Posted a ticket and potential fix to ruby-core: https://bugs.ruby-lang.org/issues/15885

deivid-rodriguez:
Even if this is fixed in core, we'll want to workaround this until the fix is included in all supported rubies, so merging.

PR #605, opened 2019-11-19
Add a setting to omit the exit prompt

### LINKED ISSUE #404: Configure `q` to quit unconditionally
## Problem description

I want `q` to always quit byebug unconditionally. I don't ever type `q` by mistake hence the added prompt, or having to append `!`, are pure overhead for me.

## Expected behavior

It would be great to have a configuration option of some kind to have `q` quit unconditionally.

## Actual behavior

Currently one has to either press q, enter, y or type `q!` (I think, never actually tried that).

## Steps to reproduce the problem

Try to quit byebug

blase:
With `no_exit_prompt` setting byebug will always quit immediately when `q` is pressed.

Resolves #404

### DIFF
--- a/lib/byebug/commands/quit.rb
+++ b/lib/byebug/commands/quit.rb
@@ -31,7 +31,7 @@ module Byebug
     end
 
     def execute
-      return unless @match[1] || confirm(pr("quit.confirmations.really"))
+      return unless @match[1] || !Setting[:prompt_on_exit] || confirm(pr("quit.confirmations.really"))
 
       processor.interface.autosave
       processor.interface.close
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Hi @blase! This looks good to me :heart:.

Only thing is whether we should call this "exit_prompt"  and default it to "true" (or maybe "prompt_on_exit", for consistency with the "stack_on_error" setting). This would be just to avoid a "negated setting" with a default value of "false", which feels a bit convoluted to me.

What do you think?

blase:
Good idea! I was changing this back and forth because I wasn't sure whether I prefer negation in the config or the `unless` clause. Defaulting to `true` and consistent naming seems like the best approach, thanks.

p:
Thank you for this PR. https://github.com/deivid-rodriguez/byebug does not mention configuration as far as I can see, how does a user configure this setting?

p-mongo:
Ping. I would like to test this patch.

PR #628, opened 2020-01-15
Readme: add new commands

rhombl4:
Update a readme file with new commands that were added in previous releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Command     | Aliases         | Subcommands
 `catch`     | `cat`           |
 `condition` | `cond`          |
 `continue`  | `c` `cont`      |
+`continue!` |                 |
 `debug`     |                 |
 `delete`    | `del`           |
 `disable`   | `dis`           | `breakpoints` `display`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,7 +124,8 @@ Command     | Aliases         | Subcommands
 `method`    | `m`             | `instance`
 `next`      | `n`             |
 `pry`       |                 |
-`quit`      | `q`             |
+`quit`      | `q`             | `!`
```
deivid-rodriguez: I guess you're adding `!` as a subcommand because `quit !` and `q !` (command followed by a space and then `!`) works?

rhombl4: Yes, it works. Also, there is subcommand `unconditionally`(with the same behavior as `!`). I can add this too.
And the same subcommands for a command `continue`.

rhombl4: It is an option for command, not a sub-command. Canceled this change.

rhombl4:
Now, these commands are looking as duplicate. But they should be present in this list, otherwise, users will not be able to learn about these helpful commands.

PR #635, opened 2020-01-23
Go back to a full `require` for requiring the extension

deivid-rodriguez:
This fixes byebug usage in Fedora. Fedora's rubygems uses a switch to prevent compiled extensions from being installed into each gem's lib/ directory (`Gem.install_extension_in_lib = false`). So byebug doesn't properly work there.

Going back to a require fixes the issue.

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -1,0 +1,34 @@
+---
+
+name: fedora
+
+on:
+  pull_request:
+
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks to @jorge-barreto for reporting this issue! :purple_heart:

conf:
Thank you for fixing this so quickly! Just stumbled upon this issue and you already fixed this ❤️

PR #648, opened 2020-02-04
runner_against_program_with_byebug_call_test: call the correct ruby

terceiro:
On Debian, while we are transitioning between Ruby versions, we have
both the old ruby (e.g. ruby2.5 and ruby2.7) and the new coinstalled,
and we want to test stuff against both of them.


deivid-rodriguez:
Thanks @terceiro! :)

PR #649, opened 2020-02-04
test: drop usage of require_relative to code in lib

terceiro:
In Debian and in other QA efforts, one might want to run the test suite
against the code that is installed on the system (as opposed to the one
in the source tree). Using require_relative like this prevents that from
working.


deivid-rodriguez:
Thanks again @terceiro!

PR #663, opened 2020-04-12
Improve Windows CI

deivid-rodriguez:
Small follow up to #662.


MSP-Greg:
Nice.  JFYI, all the recent Ruby Windows releases (2.5.8, 2.6.6, 2.7.1) include reline.  I've been working a lot on actions lately, so I haven't been following a lot of other things.  You may have already looked at reline.

deivid-rodriguez:
I know. That's why I'm explicitly sticking with the previous series for the Windows CI for now until I can have a closer look at reline.

PR #665, opened 2020-04-12
Drop ruby 2.4 support

deivid-rodriguez:
Since upstream support has ended: https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/


deivid-rodriguez:
Merged as #731.

PR #679, opened 2020-05-05
Use +where+ argument to print nth most recent frames

sergioro9:
At times the frame stack is too deep and the output of +where+ is difficult to read. Let +where+ take an argument to print the nth most recent frames.


vittorius:
@deivid-rodriguez Should we revisit this? Seems pretty useful. And I didn't find any other command or command option to replicate this functionality 😞

PR #681, opened 2020-05-05
Pass argument to +where+ to print the nth first frames

sergioro9:
At times the frame stack is too deep and the output of +where+ is difficult to read. Let +where+ take an argument to print the nth most recent frames.


deivid-rodriguez:
I like this idea! Could you please add some tests?

sergioro9:
I've added tests for `where` with an argument.

sergioro9:
Closed this pull request because I made a typo.

PR #686, opened 2020-05-26
Break in current line if `break` has no arguments

sergioro9:
Set a breakpoint at the next instruction to be executed in the selected stack frame. Like GDB.


sergioro9:
Closed this pull request because I made a typo.

PR #688, opened 2020-05-26
Pass argument to `where` to print the nth first frames

sergioro9:
When the frame stack is too deep it's useful pass an argument to `where` to print the nth most recent frames. This pull request replaces  my previous one #681 which had a typo.


deivid-rodriguez:
I'm sorry about the delay @sergioro9, hope to merge and release this soon.

PR #690, opened 2020-05-26
Break at current line when `break` has no arguments

sergioro9:
Makes it easier to create breakpoints with just `break` or `b`, like in GDB.


deivid-rodriguez (approved):
I decided to accept this. Thanks! :purple_heart:

PR #693, opened 2020-05-26
Set default history file to ~/.byebug_history

sergioro9:
Put history file in home directory. Instead of having many local
 `./.byebug_history` files, a central `~/.byebug_history` file allows to
 retreive the history independently from the current directory.


sergioro9:
Sorry I made pull request from wrong branch

PR #694, opened 2020-05-26
Set default history file to ~/.byebug_history

sergioro9:
Put history file in home directory. Instead of creating many local `./.byebug_history` files, a central `~/.byebug_history` file allows to retreive the history independently from the current directory.


deivid-rodriguez:
I'm not too sure about this, what are the advantages? A per project history file allows for a more meaningful history of commands in my opinion.

sergioro9:
I find it useful to reuse byebug commands from any directory, but this is opinion based so better to close this pull request. Thanks.

PR #727, opened 2020-09-19
Add a fallback for syntax_valid? if RubyVM is not defined

eregon:
* Use Ripper as it is a standard library.
* Another possibility would be to use eval() like this:
  ```ruby
  catch(:valid) do
    eval("BEGIN{throw :valid, true}\n#{code}")
    false
  end
  ```
  as done in https://github.com/banister/method_source/blob/836d7047ab/lib/method_source/code_helpers.rb#L66-L80
  But that means creating AST nodes, etc and seems heavier.

While trying byebug on TruffleRuby, I found this is the first issue. This change seems to fix it.

The next issue is supporting the TracePoint `:call` event in TruffleRuby: https://github.com/oracle/truffleruby/issues/1672.

### DIFF
--- a/lib/byebug/helpers/parse.rb
+++ b/lib/byebug/helpers/parse.rb
@@ -35,11 +35,18 @@ module Byebug
       def syntax_valid?(code)
         return true unless code
 
-        without_stderr do
-          RubyVM::InstructionSequence.compile(code)
-          true
-        rescue SyntaxError
-          false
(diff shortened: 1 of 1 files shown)


eregon:
The [CI failure](https://github.com/deivid-rodriguez/byebug/pull/727/checks?check_run_id=1138033932) is unrelated.

deivid-rodriguez (approved):
Thanks!

PR #747, opened 2020-10-14
Allow Breakpoint#hit_condition=(nil)

### LINKED ISSUE #739: Hit condition cannot be set to nil despite documentation
## Problem description

The documentation on `Breakpoint#hit_condition=` indicates that `nil` should be a valid value. However, the implementation doesn't actually handle `nil`.

https://github.com/deivid-rodriguez/byebug/blob/c27c0c74c67743d20d6ceae58fe242118545dc50/ext/byebug/breakpoint.c#L114-L142

## Expected behavior

Calling `bp.hit_condition=(nil)` sets `breakpoint->hit_condition = HIT_COND_NONE;`

## Actual behavior

`TypeError (nil is not a symbol)`

## Steps to reproduce the problem

```
irb> require 'byebug/core'
irb> bp = Byebug::Breakpoint.add('file.rb', 1)
irb> bp.hit_condition = nil
```

firelizzard18:
Closes #739.

This MR modifies `brkpt_set_hit_condition` in `ext/byebug/breakpoint.c` to allow clearing a breakpoint's hit condition by setting it to nil. The documentation on the function indicates this should be possible, but passing nil will raise an exception, and no code path in the function will set the condition to `HIT_COND_NONE`.

### DIFF
--- a/ext/byebug/breakpoint.c
+++ b/ext/byebug/breakpoint.c
@@ -128,6 +128,13 @@ brkpt_set_hit_condition(VALUE self, VALUE value)
   ID id_value;
 
   Data_Get_Struct(self, breakpoint_t, breakpoint);
+
+  if (NIL_P(value))
+  {
+    breakpoint->hit_condition = HIT_COND_NONE;
+    return value;
+  }
+
   id_value = rb_to_id(value);
 
   if (rb_intern("greater_or_equal") == id_value || rb_intern("ge") == id_value)


deivid-rodriguez:
Thanks @firelizzard18! Is there any user facing command to actually unset this?

deivid-rodriguez:
Actually I now see from your issue and `hit_condition` is not really exposed and can only be used programamtically. Any ideas on how to better expose it?

In any case, this looks good!

deivid-rodriguez (approved):

firelizzard18:
> Any ideas on how to better expose it?

`ConditionCommand` could be modified to support hit conditions. Possibly:

* `cond[ition] [expr[ession]] <n>[ expr]`
* `cond[ition] hit[count] <n>[ <op> value]`

Implemented as:

```ruby
```

deivid-rodriguez:
Since the code in the extension it's already there, makes sense to add it. I like your proposal :+1:.

PR #824, opened 2021-08-01
Load IRB lazily

pocke:
This PR makes loading IRB lazy. It improves startup time.

# Problem

Byebug always loads IRB, but IRB is necessary only for `irb` command. It makes startup time slower unnecessarily for non-IRB users.

# Solution

Let's load IRB lazily.

# Benchmark

About 55ms (50%) faster.

## before

```bash
$ RUBYLIB=lib ruby -rbenchmark -e 'p Benchmark.realtime{require "byebug/core"};'
0.11415267689153552
```

## after

```bash
$ RUBYLIB=lib ruby -rbenchmark -e 'p Benchmark.realtime{require "byebug/core"};'
0.059115791926160455
```

### DIFF
--- a/lib/byebug/commands/irb.rb
+++ b/lib/byebug/commands/irb.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative "../command"
-require "irb"
 require "English"
 
 module Byebug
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thank you!

PR #847, opened 2023-10-11
 Use TracePoint.allow_reentry to support TracePoint events while eval

### LINKED ISSUE #564: Incompatibility with Zeitwerk
Let me open this issue as a way to exchange impressions about this.

[Zeitwerk](https://github.com/fxn/zeitwerk) listens to `:class` events to load what the project calls "explicit namespaces" (see why at [this moment](https://youtu.be/ulCBLpCU6aY?t=2087) in my talk in RailsConf), but within a Byebug session, these events are not emitted.

Does Byebug need specifically to disable `:class` events, or could it disable others and preserve this one? If it needs them, could it be a way to make both projects compatible?

Right now, Rails 6 applications with common defaults have this gotcha, for example.

marshall-lee:
- Update Ruby and gems versions (needed to run a new test case using `TracePoint` API on new Rubies)
- Undefine alloc funcs of `Context` and `ThreadsTable`. This is to suppress a warning of this kind on modern rubies:

      warning: undefining the allocator of T_DATA class Byebug::ThreadsTable
      warning: undefining the allocator of T_DATA class Byebug::Context

- Use `TracePoint.allow_reentry` to support others' `TracePoint` events while evaling.
- Temporarily shut Byebug's `TracePoint` event handlers in the thread that issued `allowing_other_threads { ... }` to avoid infinite recursion in event handlers. `CTX_FL_IGNORE` flag was useful here.
- Add a test case using `TracePoint.new(:class)` hook inspired by @fxn's comment https://github.com/deivid-rodriguez/byebug/issues/564#issuecomment-497499214

Fixes #564


ext/byebug/breakpoint.c
```diff
@@ -370,20 +370,17 @@ check_breakpoint_by_hit_condition(VALUE rb_breakpoint)
...
       if (breakpoint->hit_count == breakpoint->hit_value)
         return 1;
       break;
     }
-    case HIT_COND_MOD:
-    {
+    case HIT_COND_MOD: {
       if (breakpoint->hit_count % breakpoint->hit_value == 0)
         return 1;
       break;
```
marshall-lee, line 386: This is not me, it's `clang-format`.

ext/byebug/context.c
```diff
@@ -658,6 +658,7 @@ void
 Init_byebug_context(VALUE mByebug)
 {
   cContext = rb_define_class_under(mByebug, "Context", rb_cObject);
+  rb_undef_alloc_func(cContext);
```
marshall-lee, line 661: Wasn't really needed but the warning is annoying.

d4rky-pl:
I have checked this PR in our codebase on some simple examples where Zeitwerk was previously failing and it seems to be working fine.

@deivid-rodriguez I know this gem has been essentially unmaintained since 2021 but is there any chance you could take a look? 🙏

douglas:
Hello @deivid-rodriguez ! Just to echo @d4rky-pl comment above, could you take a look at this?

AlexWayfer:
@deivid-rodriguez please, take a look, the issue is still here and it's annoying.

deivid-rodriguez:
Hei! Sorry. Indeed I was not able to pay a lot of attention to this library recently. I'll review this PR and make a new release in the following weeks 👍.

AlexWayfer:
@deivid-rodriguez friendly reminder. :slightly_smiling_face:

fxn:
❤️

marshall-lee:
@deivid-rodriguez THANKS

aThorp96:
Commenting for posterity incase anyone is googling around and may find this issue. This PR solved an issue I was experiencing with Rails, RSpec, Factorybot, and Zeitwerk when using `binding.pry` with byebug. The issue was if I dropped into `binding.pry` before any of the factories had been evaluated in their `let` blocks, calling the `let` fixtures would cause factorybot to start importing with `zeitwerk`, leading to a mix of `NameError: uninitialized constant` and `already initialized constant` errors. Switching the gemfile to use a version of `byebug` with this PR (specifying `gem 'byebug', github: 'deivid-rodriguez/byebug'`) fixed the issue.

I didn't setup a minimal repro but it would be something like this

```ruby
RSpec.describe MyService do
    let(:service) { create :my_service }
    
    describe "#perform" do
        let(:perform) { service.perform }
        before { perform }
        it "performs" do
            expect(service.performed).to be_true
        end
    end
```

In such an example, changing the `before` block to the following and from the pry console running `> perform` would cause a NameError due to import issues
```ruby
before do
    binding.pry
    perform
end
```

However if I instead changed the `before` block to the following and ran `> perform` the code would work as expected
```ruby
before do
    perform
    binding.pry
end
```

Issue #851, opened 2024-09-06
On Ruby 3.3.5: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0


davidrunger:
## Problem description

A warning is printed to stderr when using byebug on Ruby 3.3.5:

> /home/david/.rbenv/versions/3.3.5/lib/ruby/3.3.0/reline.rb:9: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
>
> You can add fiddle to your Gemfile or gemspec to silence this warning.

## Expected behavior

No warning is printed.

## Actual behavior

The aforementioned warning is printed.

### Reproduction script

```rb
# File: run-byebug-from-inline-gemfile.rb
# Run with: `ruby run-byebug-from-inline-gemfile.rb`

require 'bundler/inline'

gemfile do
  ruby '3.3.5'
  source 'https://rubygems.org'
  gem 'byebug', github: 'deivid-rodriguez/byebug', ref: '7e7384836defcfb40b675f5405428515b48eea9e'
end

puts("Ruby version: #{RUBY_VERSION}")

require 'byebug/version'
puts("Byebug version: #{Byebug::VERSION}")

require 'fiddle/version'
puts("Fiddle version: #{Fiddle::VERSION}")

byebug
```

### Output

```
$ ruby run-byebug-from-inline-gemfile.rb
Ruby version: 3.3.5
Byebug version: 11.1.3
Fiddle version: 1.1.2
/home/david/.rbenv/versions/3.3.5/lib/ruby/3.3.0/reline.rb:9: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add fiddle to your Gemfile or gemspec to silence this warning.
```

This issue is about the fact that those last two lines of warning output are printed. We don't want any warnings to be printed.

davidrunger:
I believe that this issue is at least somewhat related to this Ruby bug: https://bugs.ruby-lang.org/issues/20714 .

That Ruby bug, in turn, has generated these two PRs:

1. https://github.com/ruby/ruby/pull/11545
1. https://github.com/ruby/ruby/pull/11550

However:

1. I guess that, even if those PRs fix the issue in future versions of Ruby, this issue still exists (and presumably always will) in at least Ruby version 3.3.5.
[...]

PR #852, opened 2024-09-06
Add 'fiddle' gem as a runtime dependency

davidrunger:
Fixes deivid-rodriguez/byebug#851

The following warning is printed to stderr when using byebug on Ruby 3.3.5:

> /home/david/.rbenv/versions/3.3.5/lib/ruby/3.3.0/reline.rb:9: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
>
> You can add fiddle to your Gemfile or gemspec to silence this warning.

This change aims to cause that warning no longer to appear, by implementing one of the suggestions in the warning message, i.e. by adding the `fiddle` gem to byebug's gemspec as a runtime dependency.

### DIFF
--- a/byebug.gemspec
+++ b/byebug.gemspec
@@ -25,5 +25,7 @@ Gem::Specification.new do |s|
   s.extensions = ["ext/byebug/extconf.rb"]
   s.require_path = "lib"
 
+  s.add_dependency "fiddle", "~> 1.0"
+
   s.add_development_dependency "bundler", "~> 2.0"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     byebug (11.1.3)
+      fiddle (~> 1.0)
 
 GEM
   remote: https://rubygems.org/
@@ -15,6 +16,7 @@ GEM
     docile (1.3.4)
     faraday (1.0.0)
       multipart-post (>= 1.2, < 3)
+    fiddle (1.1.2)
     method_source (1.0.0)
     minitest (5.14.4)
     multipart-post (2.1.1)


davidrunger:
Here's the reproduction script from issue #851, but modified to use this branch, rather than the `master` branch of this repo:

```rb
# File: run-byebug-from-inline-gemfile.rb
# Run with: `ruby run-byebug-from-inline-gemfile.rb`

require 'bundler/inline'

gemfile do
  ruby '3.3.5'
  source 'https://rubygems.org'
  gem 'byebug', github: 'davidrunger/byebug', ref: '731e9dc15eee40bde99f46c12c68c32f09f0d806'
end

puts("Ruby version: #{RUBY_VERSION}")

require 'byebug/version'
puts("Byebug version: #{Byebug::VERSION}")

require 'fiddle/version'
puts("Fiddle version: #{Fiddle::VERSION}")

byebug
```

When executed, that modified version of the script does not print the warning about `fiddle` that we see in the issue report; it just prints this:

```
$ ruby run-byebug-from-inline-gemfile.rb
Ruby version: 3.3.5
Byebug version: 11.1.3
Fiddle version: 1.1.2
```

byebug.gemspec
```diff
@@ -25,5 +25,7 @@ Gem::Specification.new do |s|
   s.extensions = ["ext/byebug/extconf.rb"]
   s.require_path = "lib"
 
+  s.add_dependency "fiddle", "~> 1.0"
```
davidrunger, line 28: It definitely feels kind of sad to be taking `byebug` from being a dependency-free gem to now having a runtime dependency, but I guess that unfortunately that might sort of be the inevitable result for some gems of Ruby moving more and more of its non-core functionality out of the standard library and into gems.

davidrunger: I wasn't sure which version specifier to put here. I went with `~> 1.0` because:

1. `fiddle` v1.0.0 was released quite a while ago, on 2017-12-06 ([source](https://rubygems.org/gems/fiddle/versions)).
2. That ensures that we will require the latest `fiddle` major version (i.e. version 1) and that we won't pull in any subsequent major version (e.g. version 2) that might include breaking changes.

davidrunger:
# Caveat :exclamation: 

This PR should probably only be merged if `byebug` actually needs (or at least benefits from) something that is provided by the `fiddle` gem.

(Unfortunately, I personally have no idea whether or not that is the case.)

If `byebug` _does not_ need any of the functionality of `fiddle`, then it might be / probably is possible and maybe also worthwhile to figure out some other solution to #851 that does not entail adding fiddle as a not-actually-needed dependency.

Earlopain:
This warning is comming from the `reline` gem, which itself has an optional dependency on fiddle. The warning will be gone in Ruby 3.3.6 because of https://github.com/ruby/ruby/pull/11558. Also see https://github.com/ruby/reline/pull/721. Ruby decided to backport warnings for a few gems (why?)

On the other hand, `reline` should be added as a runtime dependency because you will get a warning for that with Ruby 3.4 (byebug requires `readline` but it is just an alias): https://github.com/ruby/ruby/pull/11560

Earlopain:
> FWIW, I think that this is not technically quite true. At least, in my reproduction script, I think that reline is sourced from the Ruby 3.3.5 standard library, rather than coming from the reline gem.

It is using the standard library version, yes. In ruby 3.3.5 some warnings were backported, but the standard library versions did not get appropriately updated, which means ruby's own standard library is now emitting these warnings. It is rather unfortunate, `reline` is not the only gem this happened to.


Even in the standard library I would still call it a gem, but the semantics don't really matter.

> I guess maybe it's acceptable, if 3.3.5 is the only Ruby version that will see these warnings? Maybe just close this PR and take no action to try to suppress that warning on Ruby 3.3.5

I'd love for ruby to release 3.3.6 right now but I doubt this is going to happen. There was no release when 3.3.3 broke ` net-pop` bundler resolution. Personally, I don't think libraries should work around ruby deficiencies like this, and it is just a warning. The next release will be in about ~2 months.

> Maybe it's worth mentioning that a number of my projects that use byebug actually don't ultimately print this warning

Yes, if even just one direct/transient dependency pulls it in, the warnings will not show. I don't find it optimal but it is the way it is, and to be fair it has worked rather ok for now with previous warnings. You just need one person to alert the library author.

> While pulling in the reline gem will suppress the warning about fiddle, it won't actually ensure that fiddle is made available, right?

That's correct, yes. I would think if any functionality actually relies on it, then `reline` would bump its major version instead of dropping that in a patch version. But that's just a guess, I'm as clueless as you in that regard.

davidrunger:
I seem to have unintentionally triggered the closing of this PR by deleting the branch that the PR was based on while cleaning up branches on my fork.

I guess that maybe I will just leave the PR closed, though.

One reason to leave the PR closed is that I'm not very sure that the change herein is the right change to make. I think that @Earlopain was saying that he thinks it's not the right change. He might well be right. In my fork (which is what I am using in my projects now, rather than using the upstream `byebug` gem from RubyGems or GitHub, which can result in warnings), I did ultimately [add `reline` as a dependency](https://github.com/davidrunger/runger_byebug/pull/10), as suggested by @Earlopain, rather than adding `fiddle`, as this PR does. That seems to be working well/fine.

PR #876, opened 2025-05-24
Fix warnings and on Ruby 3.4 and prepare for Ruby 3.5

pabloh:
Irb and reline are going to be available on Ruby 3.5 only as (bundled) gems, hence we need to add them as dependencies before calling require on them.

Also it's not necessary to load `irb` in memory from the get-go, since the user could be using `pry`.


cceneag:
👍

AlexWayfer:
Waiting for a new release with this! As well as `pry-byebug`.

PR #947, opened 2026-03-15
Use cross-runtime check for valid syntax

kddnewton:
Ripper may or may not go aay at some point, and we can use this code to work on all ruby runtimes without having to worry about another require.

This code itself comes directly from CRuby's test suite:

https://github.com/ruby/ruby/blob/9e89e5f3c39568add64eb4e5990575b9508089ac/tool/lib/core_assertions.rb#L157-L164

### DIFF
--- a/lib/byebug/helpers/parse.rb
+++ b/lib/byebug/helpers/parse.rb
@@ -45,10 +45,12 @@ module Byebug
             false
           end
         else
-          require "ripper" unless defined?(Ripper)
-          without_stderr do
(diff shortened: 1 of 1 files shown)


zenspider (approved):
:heart:

@deivid-rodriguez feel free to tell me to fuck off if you don't want me reviewing PRs... just trying to reduce maintenance friction.

deivid-rodriguez (approved):
Thanks both!

PR #957, opened 2026-04-15
Fix next skipping chained calls in class_eval-defined methods

### LINKED ISSUE #877: `Step` on lines with more than one thing to step into
Sometimes, I can use `step`, `next` to get to into the second call made on a line. Other times, the `next` skips the second call.

With
```
require 'byebug'
class Foo
  def some2
    true
  end
end

def some1
  Foo.new
end

byebug  
some1().some2()
"second line"
```

After `step`, `next`, I expect to be in the first line of `def some2`.

This type of stepping works for the above simple example.

```
[6, 15] in some.rb
    6:   end
    7: end
    8: 
    9: def some1
   10:   Foo.new
   11: end
   12: 
   13: byebug  
=> 14: some1().some2()
   15: "second line"
(byebug) s

[5, 14] in some.rb
    5:     true
    6:   end
    7: end
    8: 
    9: def some1
=> 10:   Foo.new
   11: end
   12: 
   13: byebug  
   14: some1().some2()
(byebug) n

[1, 10] in some.rb
    1: require 'byebug'
    2: 
    3: class Foo
    4:   def some2
=>  5:     true
    6:   end
    7: end
    8: 
    9: def some1
   10:   Foo.new
(byebug) n

[6, 15] in some.rb
    6:   end
    7: end
    8: 
    9: def some1
   10:   Foo.new
   11: end
   12: 
   13: byebug  
   14: some1().some2()
=> 15: "second line"
(byebug) n
```

Currently, I am debugging Capybara 3.40.0 and found a case where it does not happen.

```
=> 90:         check('My label')
   91:         check('My second label')
(byebug) s

[47, 56] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/dsl.rb
   47:     end
   48: 
   49:     Session::DSL_METHODS.each do |method|
   50:       class_eval <<~METHOD, __FILE__, __LINE__ + 1
   51:         def #{method}(...)
=> 52:           page.method("#{method}").call(...)
   53:         end
   54:       METHOD
   55:     end
   56:   end
(byebug) s

[41, 50] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/dsl.rb
   41:     #     end
   42:     #
   43:     # @return [Capybara::Session] The current session object
   44:     #
   45:     def page
=> 46:       Capybara.current_session
   47:     end
   48: 
   49:     Session::DSL_METHODS.each do |method|
   50:       class_eval <<~METHOD, __FILE__, __LINE__ + 1
(byebug) n

[85, 94] in some.rb
   90:         check('My label')
=> 91:         check('My second label')
(byebug) s
```

In the above, I expected to `step` into the `call(...)`, when I `next` out of the `page` call.

But instead, I have to do the below.

```
[47, 56] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/dsl.rb
   47:     end
   48: 
   49:     Session::DSL_METHODS.each do |method|
   50:       class_eval <<~METHOD, __FILE__, __LINE__ + 1
   51:         def #{method}(...)
=> 52:           page.method("#{method}").call(...)
   53:         end
   54:       METHOD
   55:     end
   56:   end
(byebug) s

[41, 50] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/dsl.rb
   41:     #     end
   42:     #
   43:     # @return [Capybara::Session] The current session object
   44:     #
   45:     def page
=> 46:       Capybara.current_session
   47:     end
   48: 
   49:     Session::DSL_METHODS.each do |method|
   50:       class_eval <<~METHOD, __FILE__, __LINE__ + 1
(byebug) s

[312, 321] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara.rb
   312:     # The current {Capybara::Session} based on what is set as {app} and {current_driver}.
   313:     #
   314:     # @return [Capybara::Session]     The currently used session
   315:     #
   316:     def current_session
=> 317:       specified_session || session_pool["#{current_driver}:#{session_name}:#{app.object_id}"]
   318:     end
   319: 
   320:     ##
   321:     #
(byebug) n

[768, 777] in /home/user/.asdf/installs/ruby/3.2.8/lib/ruby/gems/3.2.0/gems/capybara-3.40.0/lib/capybara/session.rb
   768:     end
   769: 
   770:     NODE_METHODS.each do |method|
   771:       class_eval <<~METHOD, __FILE__, __LINE__ + 1
   772:         def #{method}(...)
=> 773:           @touched = true
   774:           current_scope.#{method}(...)
   775:         end
   776:       METHOD
   777:     end
```

Now, I arrived at the method I wanted.

runephilosof-abtion:
The `next` command incorrectly skips remaining chained method calls when
inside a method defined via `class_eval` with explicit `__FILE__` and
`__LINE__` arguments (as Capybara's DSL does).

Root cause: `step_over` and `line_event` use `calced_stack_size` to
decide whether execution has returned to the target frame. This counter
tracks all frames including C calls and blocks. When a chain like
`obj.method(:foo).call(arg)` is executed, the intermediate C-level
`Method#call` inflates `calced_stack_size`, making byebug think it
hasn't returned to the original frame yet. On return, it overshoots past
the entire line.

Fix: introduce `ruby_stack_size`, a counter that only tracks Ruby method
frames (CALL/RETURN events, excluding B_CALL/B_RETURN and C events). Use
this for the `dest_frame` comparison in `step_over` and `line_event`,
so intermediate C frames don't affect the frame depth calculation.

Fixes https://github.com/deivid-rodriguez/byebug/issues/877

---
Original text

Reproduces the bug reported in issue #877: when stepping inside a method defined via class_eval with explicit __FILE__ and __LINE__ (as Capybara's DSL does), calling `next` after stepping into the first part of a chained call (e.g. get_session.method(:check).call(label)) causes byebug to skip the rest of the chain and return to the outer caller instead of staying on the same line.

The test mimics Capybara's pattern of:
  page.method("check").call(...)

and asserts that after step/step/next, the debugger should be back on the class_eval method body line (11), not the outer caller (line 24).

Ref: https://github.com/deivid-rodriguez/byebug/issues/877

### DIFF
--- a/ext/byebug/byebug.c
+++ b/ext/byebug/byebug.c
@@ -275,9 +275,9 @@ line_event(VALUE trace_point, void *data)
   if (!CTX_FL_TEST(dc, CTX_FL_IGNORE_STEPS))
     dc->steps = dc->steps <= 0 ? -1 : dc->steps - 1;
 
-  if (dc->calced_stack_size <= dc->dest_frame)
+  if (dc->ruby_stack_size <= dc->dest_frame)
   {
-    dc->dest_frame = dc->calced_stack_size;
+    dc->dest_frame = dc->ruby_stack_size;
     CTX_FL_UNSET(dc, CTX_FL_IGNORE_STEPS);
 
     dc->lines = dc->lines <= 0 ? -1 : dc->lines - 1;
@@ -309,6 +309,7 @@ call_event(VALUE trace_point, void *data)
   if (dc->calced_stack_size <= dc->dest_frame)
     CTX_FL_UNSET(dc, CTX_FL_IGNORE_STEPS);
 
+  dc->ruby_stack_size++;
   CALL_EVENT_SETUP;
 
   msym = rb_tracearg_method_id(trace_arg);
@@ -339,6 +340,8 @@ return_event(VALUE trace_point, void *data)
 
   EVENT_SETUP;
 
+  if (rb_tracearg_event_flag(trace_arg) & RUBY_EVENT_RETURN)
+    dc->ruby_stack_size--;
   RETURN_EVENT_SETUP;
 
   if ((dc->steps_out == 0) && (CTX_FL_TEST(dc, CTX_FL_STOP_ON_RET)))
--- a/ext/byebug/byebug.h
+++ b/ext/byebug/byebug.h
@@ -40,7 +40,8 @@ typedef enum
 
 typedef struct
 {
-  int calced_stack_size;
+  int calced_stack_size; /* all frames: Ruby + C + blocks          */
+  int ruby_stack_size;   /* Ruby-only frames (CALL/RETURN events)  */
   int flags;
   ctx_stop_reason stop_reason;
 
--- a/ext/byebug/context.c
+++ b/ext/byebug/context.c
@@ -87,6 +87,7 @@ byebug_context_create(VALUE thread)
 
   rb_debug_inspector_open(context_backtrace_set, (void *)context);
   context->calced_stack_size = dc_stack_size(context) + 1;
+  context->ruby_stack_size = context->calced_stack_size;
 
   if (rb_obj_class(thread) == cDebugThread)
     CTX_FL_SET(context, CTX_FL_IGNORE);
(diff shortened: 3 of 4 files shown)


runephilosof-abtion:
It was made by AI, and reviewed by me.
So I wouldn't say I understand everything, but it seems sensible.

